PR #2, opened 2019-10-25
Users

GuillaumeCisco:
Linked with:
https://github.com/SubstraFoundation/substrabac/pull/334
https://github.com/SubstraFoundation/substrafront/pull/214

Know issues:
Cannot use commands after login as no persistence has been implemented.
Several solutions:
- Use a line-oriented interpreter framework such as cmd: https://docs.python.org/3.5/library/cmd.html or https://github.com/python-cmd2/cmd2
- With SecureJWTAuthentication, store both cookies in the filesystem. How to deal with expiration? Should we use keyring?
- With SessionAuthentication, store session_id in the filesystem. How to deal with expiration?
- Login each time a command is made: quick and dirty

Solution currently used:
Store cookies in the filesystem with SecureJWTAuthentication

Improvements:
Use keyring https://github.com/jaraco/keyring for avoiding storing password in raw data. -> done
Use dedicated simple TokenAuthentification with substrabac -> done


Jenkinsfile
```diff
@@ -0,0 +1,70 @@
+pipeline {
```
Kelvin-M: We have moved to travis.yaml, so Jenkinsfile should be removed ?
@ClementGautier  @samlesu

ClementGautier: Yup, you should probably rebase (it's not present on dev anymore)

PR #4, opened 2019-10-28
Remove the tempfile module

maeldebon:
Removing the tempfile module in tests/test_cli.py

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -14,7 +14,6 @@
 
 import json
 import re
-import tempfile
 from unittest import mock
 
 import click
@@ -41,10 +40,10 @@ def execute(command, exit_code=0):
     return result.output
 
 
-def client_execute(tmpdir, command, exit_code=0):
+def client_execute(directory, command, exit_code=0):
     # force using a new config file and a new profile
     if '--config' not in command:
-        cfgpath = tmpdir / 'substra.cfg'
+        cfgpath = directory / 'substra.cfg'
         substra.sdk.config.Manager(str(cfgpath)).add_profile(
             'default', url='http://foo')
         command.extend(['--config', str(cfgpath)])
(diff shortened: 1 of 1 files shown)


tests/test_cli.py
```diff
@@ -115,57 +114,60 @@ def test_command_list_node(workdir, mocker):
     ('traintuple', ['--objective-key', 'foo', '--algo-key', 'foo', '--dataset-key', 'foo',
                     '--data-samples-path']),
     ('testtuple', ['--traintuple-key', 'foo', '--data-samples-path'])]
-)
+                         )
```
samlesu: This change does not seem related to the PR.

tests/test_cli.py
```diff
@@ -115,57 +114,60 @@ def test_command_list_node(workdir, mocker):
...
+        json_file = workdir / "temp_file.json"
+        with open(str(json_file), 'w') as f:
+            json.dump({}, f)
+        client_execute(workdir, ['add', asset_name] + params + [str(json_file)])
     assert m.is_called()
 
-    with tempfile.NamedTemporaryFile(suffix='.md') as file:
-        res = client_execute(workdir, ['add', asset_name] + params + [file.name], exit_code=2)
-        assert re.search(r'File ".*" is not a valid JSON file\.', res)
+    file = workdir / "temp_file.txt"
```
samlesu: From my point of view, we could use a better name for this variable (for instance `invalid_json_file`).
By reading the code, we would directly understand that the first test is done with a valid file and that we repeat it with an invalid file.

tests/test_cli.py
```diff
@@ -115,57 +114,60 @@ def test_command_list_node(workdir, mocker):
...
-    with tempfile.NamedTemporaryFile(mode='w', suffix='.json') as file,\
-            mock_client_call(mocker, method_name, response={}) as m:
-        json.dump({}, file)
-        file.seek(0)
-
-        client_execute(workdir, ['add', asset_name] + params + [file.name])
+    with mock_client_call(mocker, method_name, response={}) as m:
+        json_file = workdir / "temp_file.json"
+        with open(str(json_file), 'w') as f:
+            json.dump({}, f)
```
samlesu: Minor, to take full advantage of the Pathlib library, could be simplified:
```python
json_file.write_text(json.dumps({}))
```

(it's already done like this line 128)

Same comment for next tests

tests/test_cli.py
maeldebon: Yes, seems to be a mistake

tests/test_cli.py
maeldebon: Yes, I changed that for the next commit 👌

tests/test_cli.py
maeldebon: Yes, but JSON is different on this point, I already tried, unsuccessfully 🤔

samlesu: The main difference is to use `json.dumps(obj)`: it will return a JSON formatted string containing the serialized input `obj`.

You can run the following snippet in a python script:
```python
import pathlib
import json
f = pathlib.PosixPath("/tmp/foo")
f.write_text(json.dumps({}))
```
You can then check that the file `tmp/foo` exists and contains the expecting value.

PR #12, opened 2019-10-31
Automaticaly deploy to pypi.org

ClementGautier:


README.md
```diff
@@ -4,10 +4,18 @@ CLI and SDK for interacting with Substra platform.
...
+  - [Install](#install)
+  - [Usage](#usage)
+    - [CLI](#cli)
+    - [SDK](#sdk)
+  - [Documentation](#documentation)
+  - [Examples](#examples)
+  - [Contributing](#contributing)
+    - [Setup](#setup)
+    - [Documentation](#documentation-1)
+    - [Deploy](#deploy)
```
samlesu: This change seems not related to the PR scope and I'm not sure it's good idea to:
- nest all items inside `Substra`
- add the table of contents
- expand the section `Usage` and `Contributing`

LICENCE
```diff
@@ -1,190 +0,0 @@
-                                 Apache License
```
samlesu: @inalgnu can you confirm that we can safely remove this LICENCE?

inalgnu: yes! 
My bad, it was a wrong file name

README.md
ClementGautier: This change was automatically performed by my IDE to be honest. I pushed it because it updated a table of content that existed but was incomplete. I will undo the change so it stay simple

Kelvin-M:
Nice to note that we use an encrypted token as suggested  in https://docs.travis-ci.com/user/deployment/pypi/
Thank you Clément !

PR #21, opened 2019-11-05
improve documentation for add_data_sample

### LINKED ISSUE #20: Docs: add guide to ingest data in substra
This step will often be complicated and more documentation will help our users.

Questions to be answered:
- What are the modes available to add data to substra ?
- How to choose between the different modes ?
- How to add "a lot of" data samples (>100-1000) ?
- How to add data samples depending on their size (>10-100mo) ?
- ...

It would be good to:
- enrich the current documentation for the `add_data_sample(s)` command / method.
- add a dedicated guide: "How to ingest data in Substra"

samlesu:
Fix #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Learning about the Substra platform:
 - [Concepts](./docs/concepts.md)
 - [Machine Learning tasks](./docs/ml_tasks.md)
 - [Adding a full pipeline](./docs/full_pipeline_workflow.md)
+- [Add data samples](./docs/add_data_samples.md)
 
 ## Examples
 
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ Learning about the Substra platform:
 - [Concepts](./docs/concepts.md)
 - [Machine Learning tasks](./docs/ml_tasks.md)
 - [Adding a full pipeline](./docs/full_pipeline_workflow.md)
-- [Add data samples](./docs/add_data_samples.md)
+- [Adding data samples](./docs/add_data_samples.md)
 
 ## Examples
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,6 +4,7 @@ Interacting with the Substra platform:
 
 - [Command line interface](../references/cli.md)
 - [SDK](../references/sdk.md)
+- [Ingest data samples](./ingest_data.md)
 
 Implementing your assets in python:
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -16,4 +16,4 @@ Learning about the Substra platform:
 - [Concepts](./concepts.md)
 - [Machine Learning tasks](./ml_tasks.md)
 - [Adding a full pipeline](./full_pipeline_workflow.md)
-- [Add data samples](./add_data_samples.md)
+- [Adding data samples](./add_data_samples.md)
--- a/docs/README.md
+++ b/docs/README.md
@@ -4,7 +4,6 @@ Interacting with the Substra platform:
 
 - [Command line interface](../references/cli.md)
 - [SDK](../references/sdk.md)
-- [Ingest data samples](./ingest_data.md)
 
 Implementing your assets in python:
 
@@ -17,3 +16,4 @@ Learning about the Substra platform:
 - [Concepts](./concepts.md)
 - [Machine Learning tasks](./ml_tasks.md)
 - [Adding a full pipeline](./full_pipeline_workflow.md)
+- [Add data samples](./add_data_samples.md)
--- a/docs/add_data_samples.md
+++ b/docs/add_data_samples.md
@@ -1,6 +1,6 @@
-# Ingest data samples to Substra
+# Add data samples to Substra
 
-Data samples should be ingested to Substra using our python SDK or our command line
+Data samples can be ingested to Substra using our python SDK or our command line
 interface tool.
 
 2 different modes are available when adding data samples:
--- a/docs/add_data_samples.md
+++ b/docs/add_data_samples.md
@@ -8,5 +8,6 @@ interface tool.
   requests. This mode is most likely to be used to add relatively small files or for
   testing purposes.
 - remote: data are located on the server filesystem. Data must have been transferred
-  previously to the server. This mode is recommended for adding relatively big data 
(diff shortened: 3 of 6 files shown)


docs/README.md
```diff
@@ -4,6 +4,7 @@ Interacting with the Substra platform:
 
 - [Command line interface](../references/cli.md)
 - [SDK](../references/sdk.md)
+- [Ingest data samples](./ingest_data.md)
```
jmorel: I feel this link should be under "Learning about the substra platform".

It should also be included in the repo's main README.

docs/ingest_data.md
```diff
@@ -0,0 +1,12 @@
+# Ingest data samples to Substra
```
jmorel: Since we're already using "add" everywhere when talking about all assets (included data samples), we should stick with it. 

I'm not against using ingest to avoid repetition but document names, titles and introductions should use "add" for clarity.

docs/ingest_data.md
```diff
@@ -0,0 +1,12 @@
...
+
+Data samples should be ingested to Substra using our python SDK or our command line
```
jmorel: Are there any other ways than cli/sdk? If not then we should use "can" instead of "should".

references/sdk.md
```diff
@@ -48,10 +48,21 @@ Create new data sample asset.
...
+The `path` in the data dictionary must point to a directory representing the
+data sample content. Note that the directory can contain multiple files, all the
+directory content will be added to the platform.
+
+If `local` is true, `path` must refer to a directory located on the local
+filesystem. The file content will be transferred to the server through an
+HTTP query, so this mode should be used for relatively small files (<10mo).
+
+If `local` is false, `path` must refer to a directory located on the server
+filesystem. This mode is well suited for all kind of file sizes.
```
jmorel: Could you adapt this comment for the add_data_samples section as well?

docs/README.md
samlesu: Yes, thanks.

docs/ingest_data.md
samlesu: Yes, thanks.

docs/ingest_data.md
samlesu: If there are other ways, I don't think we want to document them.
I will update "should" to "can".

samlesu:
@jmorel Thanks for the review!

README.md
```diff
@@ -71,6 +71,7 @@ Learning about the Substra platform:
 - [Concepts](./docs/concepts.md)
 - [Machine Learning tasks](./docs/ml_tasks.md)
 - [Adding a full pipeline](./docs/full_pipeline_workflow.md)
+- [Add data samples](./docs/add_data_samples.md)
```
jmorel: To keep in line with the existing wording we could use "Adding data samples"

references/sdk.md
```diff
@@ -67,6 +78,13 @@ Create many data sample assets.
     "test_only": bool,
 }
 ```
+Create multiple data samples through a single HTTP request.
+
+For the `local` argument, please refer to the method `Client.add_data_sample`.
```
jmorel: It's not exactly the same though, the structure changes for data_samples (plural): the path must point to a directory containing only subdirectories, each of which will be turned into a data sample.

references/sdk.md
samlesu: Yes, thanks.

references/sdk.md
samlesu: Yes, your comment applies only to the `path` argument. Is that correct?
I think the `local` argument has exactly the same meaning when adding one or many data samples.

jmorel: Absolutely, sorry. You're right, local has the same meaning it's path that changes.

samlesu: I added some additional info for the `paths` field.
It's only in the CLI that the `path` option can point to a directory containing subdirectories with the option `multiple` set to True.

docs/add_data_samples.md
```diff
@@ -0,0 +1,12 @@
...
+
+Data samples can be ingested to Substra using our python SDK or our command line
+interface tool.
+
+2 different modes are available when adding data samples:
+- local: data are located on the local filesystem and are transferred through HTTP
+  requests. This mode is most likely to be used to add relatively small files or for
+  testing purposes.
+- remote: data are located on the server filesystem. Data must have been transferred
+  previously to the server. This mode is recommended for adding relatively big data
```
Kelvin-M: `Data must have been transferred previously to the server.` could not be clear for the user. You can maybe develop this part ?

references/sdk.md
Kelvin-M: It should also be accessible (readable) for `substra-backend`

substra/sdk/client.py
```diff
@@ -164,9 +163,20 @@ def add_data_sample(self, data, local=True, timeout=False,
...
+        The `path` in the data dictionary must point to a directory representing the
+        data sample content. Note that the directory can contain multiple files, all the
+        directory content will be added to the platform.
+
+        If `local` is true, `path` must refer to a directory located on the local
+        filesystem. The file content will be transferred to the server through an
+        HTTP query, so this mode should be used for relatively small files (<10mo).
+
+        If `local` is false, `path` must refer to a directory located on the server
+        filesystem. This mode is well suited for all kind of file sizes.
```
Kelvin-M: Same as previous comment, the `path` should be accessible (readable) for substra-backend.

docs/add_data_samples.md
samlesu: Yes, good idea. Do you have some suggestions to develop this part?

references/sdk.md
samlesu: Done.

substra/sdk/client.py
samlesu: Done

docs/add_data_samples.md
Kelvin-M: Something like you may need to ask the administrator of the substra node how to transfer your data. It could be ssh ou sftp or another protocol

samlesu: Thanks. Done.

Issue #6, opened 2019-10-29
CLI: add in model keys for add_traintuple command 


samlesu:
It will be great to add this missing optional parameter.
A `TODO` already exists in the code and should be removed.

maeldebon:
Done in #29

PR #29, opened 2019-11-12
Add in model keys to add_traintuple - Issue #6

maeldebon:
Related issue: #6 
Adding an optional parameter named in_models_keys


substra/cli/interface.py
```diff
@@ -445,6 +445,7 @@ def add_algo(ctx, data, output_format, config, profile, user, verbose):
 @click.option('--data-samples-path', 'data_samples', required=True,
               type=click.Path(exists=True, resolve_path=True, dir_okay=False),
               callback=load_json_from_path)
+@click.option('--in-models-keys')
```
jmorel: Could add more details about this option? The type for example, whether it's multiple or not etc.

Issue #31, opened 2019-11-13
CLI: add permissions option


jmorel:
When adding an asset through the CLI, there is currently no way to set permissions, they are always set to the default "public" value.

We need to design a way to add permissions through options and implement it.

jmorel:
I just realized that we already discussed what such an option could be like in #5 .

The chose method is to have an `--permissions-path` option (not mandatory) that accepts as input a path to a json file whose structure is:
```json
{
    "permissions": {
        "public": bool,
        "authorizedIDs": list[str],
    }
}
```

If not specified, the default "public" permissions will be used.

jmorel:
So... I was mistaken. Quite badly really. The PATH/data argument used for add objective/dataset/algo already allows to pass "permissions". So there's really no need for it. 

Sorry @maeldebon, this was all for nothing.

PR #32, opened 2019-11-13
Add permissions option

maeldebon:
Related issue: #31 
Add a --permissions-path option

### DIFF
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -310,7 +310,8 @@ def add_data_sample(ctx, path, dataset_key, local, multiple, test_only,
 @click_option_verbose
 @click.pass_context
 @error_printer
-def add_dataset(ctx, data, objective_key, output_format, config, permissions, profile, user, verbose):
+def add_dataset(ctx, data, objective_key, output_format, config, permissions, profile, user,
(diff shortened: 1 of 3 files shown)


substra/cli/interface.py
```diff
@@ -148,6 +148,17 @@ def click_option_verbose(f):
     )(f)
 
 
+def click_option_permissions(f):
+    """Add permissions file path to command."""
+    return click.option(
+        '--permissions-path',
+        'permissions',
+        type=click.Path(exists=True, dir_okay=False), callback=load_json_from_path,
```
jmorel: Could you move callback to its own line just to make it clear it's not part of the type?

substra/cli/interface.py
```diff
@@ -148,6 +148,17 @@ def click_option_verbose(f):
...
+        metavar="PATH",
```
jmorel: I'm not sure but I think metavar only work with arguments, not options, could you double check against the documentation?

maeldebon: Indeed, it was a mistake

maeldebon:
As written in #31, this PR is not required

PR #34, opened 2019-11-14
Add aggregate algo & traintuple

maeldebon:
- [x] implement methods `add`/`list`/`get` composite algo in CLI/SDK
- [x] implement methods `add`/`list`/`get` composite traintuple in CLI/SDK
_____________

Associated PRs:
- [backend](https://github.com/SubstraFoundation/substra-backend/pull/40)
- [chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/26)
- [tools](https://github.com/SubstraFoundation/substra-tools/pull/9)

### DIFF
--- a/references/cli.md
+++ b/references/cli.md
@@ -7,8 +7,10 @@
 - [substra add objective](#substra-add-objective)
 - [substra add algo](#substra-add-algo)
 - [substra add compute_plan](#substra-add-compute_plan)
+- [substra add aggregate_algo](#substra-add-aggregate_algo)
 - [substra add composite_algo](#substra-add-composite_algo)
 - [substra add traintuple](#substra-add-traintuple)
+- [substra add aggregatetuple](#substra-add-aggregatetuple)
 - [substra add composite_traintuple](#substra-add-composite_traintuple)
 - [substra add testtuple](#substra-add-testtuple)
 - [substra get](#substra-get)
@@ -245,6 +247,43 @@ Options:
   --help                Show this message and exit.
 ```
 
+## substra add aggregate_algo
+
+```bash
+Usage: substra add aggregate_algo [OPTIONS] PATH
+
+  Add aggregate algo.
+
+  The path must point to a valid JSON file with the following schema:
+
+  {
+      "name": str,
+      "description": path,
+      "file": path,
+      "permissions": {
+          "public": bool,
+          "authorized_ids": list[str],
+      },
+  }
+
+  Where:
+  - name: name of the algorithm
+  - description: path to a markdown file describing the algo
+  - file: path to tar.gz or zip archive containing the algorithm python
+    script and its Dockerfile
+  - permissions: define asset access permissions
+
+Options:
+  --yaml          Display output as yaml.
+  --json          Display output as json.
+  --pretty        Pretty print output  [default: True]
+  --config PATH   Config path (default ~/.substra).
+  --profile TEXT  Profile name to use.
+  --user FILE     User file path to use (default ~/.substra-user).
+  --verbose       Enable verbose mode.
+  --help          Show this message and exit.
+```
+
 ## substra add composite_algo
 
 ```bash
@@ -314,6 +353,30 @@ Options:
   --user FILE               User file path to use (default ~/.substra-user).
   --verbose                 Enable verbose mode.
   --help                    Show this message and exit.
+```
+
+## substra add aggregatetuple
+
+```bash
+Usage: substra add aggregatetuple [OPTIONS]
+
+  Add aggregatetuple.
+
+Options:
+  --objective-key TEXT  [required]
+  --algo-key TEXT       [required]
+  --in-model-key TEXT   In model traintuple key.
+  --worker TEXT         Node ID for worker execution.  [required]
+  --rank INTEGER
+  --tag TEXT
+  --yaml                Display output as yaml.
+  --json                Display output as json.
+  --pretty              Pretty print output  [default: True]
+  --config PATH         Config path (default ~/.substra).
+  --profile TEXT        Profile name to use.
+  --user FILE           User file path to use (default ~/.substra-user).
+  --verbose             Enable verbose mode.
+  --help                Show this message and exit.
 ```
 
 ## substra add composite_traintuple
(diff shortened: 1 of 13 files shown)


substra/sdk/client.py
```diff
@@ -350,6 +380,31 @@ def add_traintuple(self, data, timeout=False, exist_ok=False):
...
+
+        `data` is a dict object with the following schema:
+
+```
+        {
+            "algo_key": str,
+            "objective_key": str,
+            "data_manager_key": str,
+            "tag": str,
+            "compute_plan_id": str,
```
samlesu: I think this is not compliant with the specification.
The `data_manager_key` is not needed.
The following fields are missing:
- worker
- in_model_ids
- rank

substra/cli/interface.py
```diff
@@ -543,29 +546,25 @@ def add_traintuple(ctx, objective_key, algo_key, dataset_key, data_samples, in_m
...
-        'data_manager_key': dataset_key,
     }
 
+    if in_models_keys:
+        data['in_models_keys'] = in_models_keys
+
+    if worker:
+        data['worker'] = worker
+
+    if rank:
```
samlesu: If `rank` is equal to 0, the rank won't be set. In this particular case I think we should check it is not equal to None (that will be the case if rank is not provided).

substra/cli/interface.py
```diff
@@ -491,6 +531,47 @@ def add_traintuple(ctx, objective_key, algo_key, dataset_key, data_samples, in_m
     printer.print(res, is_list=False)
 
 
+@add.command('aggregate_traintuple')
+@click.option('--objective-key', required=True)
+@click.option('--algo-key', required=True)
+@click.option('--in-model-key', 'in_models_keys', type=click.STRING, multiple=True,
+              help='In model traintuple key.')
+@click.option('--worker')
```
jmorel: Could check whether the worker is mandatory or not? The spec isn't very clear about this.

substra/cli/interface.py
```diff
@@ -491,6 +531,47 @@ def add_traintuple(ctx, objective_key, algo_key, dataset_key, data_samples, in_m
     printer.print(res, is_list=False)
 
 
+@add.command('aggregate_traintuple')
+@click.option('--objective-key', required=True)
+@click.option('--algo-key', required=True)
+@click.option('--in-model-key', 'in_models_keys', type=click.STRING, multiple=True,
```
jmorel: For an aggregate traintuple, these in model keys are mandatory I think.

substra/cli/interface.py
samlesu: We can start by making it mandatory. 

We could make it optional later by using the creator as the worker? If yes, it could be done by the chaincode directly.

substra/cli/interface.py
samlesu: I'm not sure we should force input models. We could create an aggregate without input models (kind of initialization step).

references/cli.md
```diff
@@ -316,6 +355,30 @@ Options:
...
+## substra add aggregatetuple
+
+```bash
+Usage: substra add aggregatetuple [OPTIONS]
+
+  Add aggregatetuple.
+
+Options:
+  --objective-key TEXT  [required]
+  --algo-key TEXT       [required]
```
jmorel: Shouldn't this be `--aggregate-algo-key`?

references/cli.md
```diff
@@ -316,6 +355,30 @@ Options:
...
+  --in-model-key TEXT   In model traintuple key.
```
jmorel: This is a "multiple" option right? Could you mention it in the doc?

references/cli.md
```diff
@@ -316,6 +355,30 @@ Options:
...
+  --rank INTEGER
+  --tag TEXT
+  --yaml                Display output as yaml.
+  --json                Display output as json.
+  --pretty              Pretty print output  [default: True]
+  --config PATH         Config path (default ~/.substra).
+  --profile TEXT        Profile name to use.
+  --user FILE           User file path to use (default ~/.substra-user).
+  --verbose             Enable verbose mode.
+  --help                Show this message and exit.
```
jmorel: Isn't the `--compute-plan-id` option missing here?

references/sdk.md
```diff
@@ -183,7 +201,6 @@ Create new composite algo asset.
     },
 }
 ```
-
```
jmorel: Since this is markdown shouldn't we keep these empty lines?

references/sdk.md
```diff
@@ -210,14 +227,32 @@ Create new traintuple asset.
...
 
+## add_aggregatetuple
+```python
+Client.add_aggregatetuple(self, data, timeout=False, exist_ok=False)
+```
+Create new aggregatetuple asset.
+`data` is a dict object with the following schema:
+```
+{
+    "algo_key": str,
```
jmorel: Same remark as for the cli regarding using `aggregate_algo_key` instead of `algo_key`

substra/cli/interface.py
```diff
@@ -580,6 +620,47 @@ def add_traintuple(ctx, objective_key, algo_key, dataset_key, data_samples, in_m
...
+    data = {
+        'algo_key': algo_key,
+        'objective_key': objective_key,
+    }
+
+    if in_models_keys:
+        data['in_models_keys'] = in_models_keys
+
+    if worker:
+        data['worker'] = worker
```
jmorel: Since the option is required, we can remove the if statement.

substra/cli/printers.py
```diff
@@ -349,6 +353,35 @@ class TraintuplePrinter(AssetPrinter):
...
+        Field('Status', 'status'),
+        Field('Perf', 'dataset.perf'),
+        Field('Dataset key', 'dataset.openerHash'),
+        Field('Rank', 'rank'),
+        Field('Compute Plan Id', 'computePlanID'),
+        Field('Tag', 'tag'),
+        Field('Log', 'log'),
+        Field('Creator', 'creator'),
+        Field('Worker', 'dataset.worker'),
+        PermissionField('Permissions', 'permissions'),
```
jmorel: Could you add the inModel keys as well?

PR #37, opened 2019-11-14
Remove reference to private docker registry

jmorel:
Since our docker images are now available on the public docker registry, the references to our private one are irrelevant. This PR replaces them with the new public ones.

Also, the MacOS Catalina update forced us to update the default working dir of the substra platform in `/tmp/substra`, I updated the guide accordingly.

### DIFF
--- a/docs/local_install.md
+++ b/docs/local_install.md
@@ -14,7 +14,7 @@ This guide will help you to get Substra source code and to start a local instanc
 
 - docker:
   - on macOS, by default, docker has access only to the user directory.
-    - Substra requires access to `/tmp/substra`, update accordingly the docker desktop configuration (`Preferences` > `File Sharing`).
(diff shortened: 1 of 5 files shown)


docs/local_install.md
```diff
@@ -14,7 +14,7 @@ This guide will help you to get Substra source code and to start a local instanc
 
 - docker:
   - on macOS, by default, docker has access only to the user directory.
-    - Substra requires access to `/substra`, update accordingly the docker desktop configuration (`Preferences` > `File Sharing`).
+    - Substra requires access to `/tmp/substra`, update accordingly the docker desktop configuration (`Preferences` > `File Sharing`).
```
samlesu: I think the path depends on the value of the env var `SUBSTRA_PATH` (at least for the setup with docker-compose).
Could be good to add a bit more information here.

jmorel: Done!

Issue #28, opened 2019-11-08
CLI: add profile in commands suggested by printers


jmorel:
A few of the asset printers provide commands to the user:
* a command to display the description of an asset after a get
* a command to download an asset after a get
* a command to display a leaderboard

These suggested commands currently do not include the profile that was used to run the previous command. This prevent users from simply copy/pasting the suggested commands in their terminal.

# Example

## Current ouput

```shell
1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3 --profile node-1
KEY                         1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3
NAME                        Skin Lesion Classification Objective
METRICS                     macro-average recall
TEST DATASET KEY            None
TEST DATA SAMPLE KEYS       None
OWNER                       MyPeer1MSP
PERMISSIONS                 Processable by anyone

Download this objective's metric:
	substra download objective 1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3

Display this objective's description:
	substra describe objective 1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3

Display this objective's leaderboard:
	substra leaderboard 1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3

```

## Desired output
```shell
$ substra get objective 1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3 --profile node-1
KEY                         1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3
NAME                        Skin Lesion Classification Objective
METRICS                     macro-average recall
TEST DATASET KEY            None
TEST DATA SAMPLE KEYS       None
OWNER                       MyPeer1MSP
PERMISSIONS                 Processable by anyone

Download this objective's metric:
	substra download objective 1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3 --profile node-1

Display this objective's description:
	substra describe objective 1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3 --profile node-1

Display this objective's leaderboard:
	substra leaderboard 1cdafbb018dd195690111d74916b76c96892d897ec3587c814f287946db446c3 --profile node-1
```

maeldebon:
Done!

PR #39, opened 2019-11-15
Add profile option in suggested commands - Issue #28

maeldebon:
Add the --profile option in the commands suggested by the printers
See #28

### DIFF
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -565,6 +565,7 @@ def get(ctx, asset_name, asset_key, expand, output_format, config, profile, user
     # method must exist in sdk
     method = getattr(client, f'get_{asset_name.lower()}')
     res = method(asset_key)
+    res['profile'] = profile
     printer = printers.get_asset_printer(asset_name, output_format)
     printer.print(res, expand=expand)
 
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -565,9 +565,8 @@ def get(ctx, asset_name, asset_key, expand, output_format, config, profile, user
     # method must exist in sdk
     method = getattr(client, f'get_{asset_name.lower()}')
     res = method(asset_key)
-    res['profile'] = profile
     printer = printers.get_asset_printer(asset_name, output_format)
-    printer.print(res, expand=expand)
+    printer.print(res, profile=profile, expand=expand)
 
 
 @cli.command('list')
--- a/substra/cli/printers.py
+++ b/substra/cli/printers.py
@@ -167,19 +167,24 @@ class AssetPrinter(BasePrinter):
     def _get_single_fields(self):
         return (self.key_field, ) + self.single_fields
 
+    def print_with_profile(self, profile):
+        if profile and profile != 'default':
+            return f'--profile {profile}'
+        return ''
+
     def print_download_message(self, item, profile=None):
         if self.download_message:
             key_value = self.key_field.get_value(item)
-            print()
-            print(self.download_message)
-            print(f'\tsubstra download {self.asset_name} {key_value} --profile {profile}')
+            profile_arg = self.print_with_profile(profile)
+            print('\n' + self.download_message)
+            print(f'\tsubstra download {self.asset_name} {key_value} {profile_arg}')
 
     def print_description_message(self, item, profile=None):
         if self.has_description:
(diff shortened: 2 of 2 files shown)


substra/cli/interface.py
```diff
@@ -565,6 +565,7 @@ def get(ctx, asset_name, asset_key, expand, output_format, config, profile, user
     # method must exist in sdk
     method = getattr(client, f'get_{asset_name.lower()}')
     res = method(asset_key)
+    res['profile'] = profile
```
samlesu: I would recommend to not add the profile to the results returned by the SDK / server. 

It is not part of the data we want to display to the user, it is more something related to how this should be displayed.

This could be added as an option the print command?

maeldebon: I changed that in the last commit, that's better indeed! 🙂

substra/cli/printers.py
```diff
@@ -167,30 +167,35 @@ def _get_list_fields(self):
     def _get_single_fields(self):
         return (self.key_field, ) + self.single_fields
 
-    def print_download_message(self, item):
+    def print_with_profile(self, profile):
```
jmorel: In this module, all methods starting with `print_` actually print something. You could rename it to `get_profile_arg` to make its purpose crystal clear.

substra/cli/printers.py
```diff
@@ -167,30 +167,35 @@ def _get_list_fields(self):
...
+        return ''
+
+    def print_download_message(self, item, profile=None):
         if self.download_message:
             key_value = self.key_field.get_value(item)
-            print()
-            print(self.download_message)
-            print(f'\tsubstra download {self.asset_name} {key_value}')
+            profile_arg = self.print_with_profile(profile)
+            print('\n' + self.download_message)
```
jmorel: Why the change from `print()` to `print("\n")`? It's almost the same, but now we have sometimes `\n` and sometimes `print()` in the same file.

substra/cli/printers.py
```diff
@@ -167,30 +167,35 @@ def _get_list_fields(self):
...
 
-    def print_description_message(self, item):
+    def print_description_message(self, item, profile=None):
         if self.has_description:
             key_value = self.key_field.get_value(item)
-            print()
-            print(f'Display this {self.asset_name}\'s description:')
-            print(f'\tsubstra describe {self.asset_name} {key_value}')
+            profile_arg = self.print_with_profile(profile)
+            print('\n' + f'Display this {self.asset_name}\'s description:')
```
jmorel: No need to concatenate strings here, you can write `f'\nDisplay etc.'`.

substra/cli/printers.py
```diff
@@ -237,15 +242,20 @@ class ObjectivePrinter(AssetPrinter):
     )
     download_message = 'Download this objective\'s metric:'
 
-    def print_leaderboard_message(self, item):
+    def print_with_profile(self, profile):
+        if profile and profile != 'default':
+            return f'--profile {profile}'
+        return ''
```
jmorel: Since the `AlgoPrinter` class inherits from `AssetPrinter` it automatically gets the `print_with_profile` method, no need to override it.

substra/cli/printers.py
jmorel: Another solution would be to include the `\n` at the beginning of the download_message string.

substra/cli/printers.py
maeldebon: That's true, I'll change that! 🙂

substra/cli/printers.py
maeldebon: I think that this is cleaner this way, instead of having an empty print. Adding the `\n` at the beginning of the download_message string means that you have to do that for all of these messages. So for me this is easier (and maybe cleaner) to just put only one `\n` when the message is displayed 😃

substra/cli/printers.py
maeldebon: Yes, I'll change that

substra/cli/printers.py
maeldebon: That's true, I remove it! 🙂

substra/cli/printers.py
jmorel: Sure why not. All my arguments were going to be about taste and not exactly productive. Let's go with it!

PR #40, opened 2019-11-18
Update command to update /etc/host via bash.

ClairePhi:
Manual update is suggested if both the initial way and the way using bash do not work.

### DIFF
--- a/docs/local_install_skaffold.md
+++ b/docs/local_install_skaffold.md
@@ -102,7 +102,7 @@ The last step will be to update `/etc/hosts` by adding this new line to it:
 192.168.65.2 substra-backend.node-1.com substra-frontend.node-1.com substra-backend.node-2.com substra-frontend.node-2.com
 ```
 
-Running the following command will do it for you (you 'll be asked for your password again):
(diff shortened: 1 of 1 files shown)


jmorel:
@ClairePhi Good catch, thanks for the updated command!

Also I tested your command on zsh and it works fine there as well. You could simply replace the old command with the new one, no need to mention bash.

Could you rebase your PR?

ClairePhi:
Thank you @jmorel! Done :)

docs/local_install_skaffold.md
```diff
@@ -102,10 +102,10 @@ The last step will be to update `/etc/hosts` by adding this new line to it:
 192.168.65.2 substra-backend.node-1.com substra-frontend.node-1.com substra-backend.node-2.com substra-frontend.node-2.com
 ```
 
-Running the following command will do it for you (you'll be asked for your password again)
+Running the following command will do it for you (you 'll be asked for your password again):
```
jmorel: There's an extra space here, could you remove it before the merge?

ClairePhi:
Extra space removed. Thanks @jmorel!

PR #41, opened 2019-11-21
Update usage of mocker

jmorel:
Since the last release of pytest-mock, mocker.patch isn't compatible with context managers anymore : https://github.com/pytest-dev/pytest-mock#note-about-usage-as-context-manager

This made the tests in test_runner.py fail.

This PR updates them to follow the proper usage of mocker.

See also https://github.com/pytest-dev/pytest-mock/releases/tag/v1.12.0


jmorel:
@samlesu Done!

PR #44, opened 2019-11-21
Group config, profile, user & verbose options in a global decorator

maeldebon:
As written in the following comment:
```
#  TODO profile, config, json and verbose options should be handled in a single
#      decorator to populate a GlobalOption object stored in the context
```

### DIFF
--- a/references/cli.md
+++ b/references/cli.md
@@ -110,13 +110,13 @@ Usage: substra add dataset [OPTIONS] PATH
 
 Options:
   --objective-key TEXT
-  --yaml                Display output as yaml.
-  --json                Display output as json.
-  --pretty              Pretty print output  [default: True]
   --config PATH         Config path (default ~/.substra).
   --profile TEXT        Profile name to use.
   --user FILE           User file path to use (default ~/.substra-user).
   --verbose             Enable verbose mode.
+  --yaml                Display output as yaml.
+  --json                Display output as json.
+  --pretty              Pretty print output  [default: True]
   --help                Show this message and exit.
 ```
 
@@ -161,13 +161,13 @@ Usage: substra add objective [OPTIONS] PATH
 Options:
   --dataset-key TEXT
   --data-samples-path FILE  Test data samples.
-  --yaml                    Display output as yaml.
-  --json                    Display output as json.
-  --pretty                  Pretty print output  [default: True]
   --config PATH             Config path (default ~/.substra).
   --profile TEXT            Profile name to use.
   --user FILE               User file path to use (default ~/.substra-user).
   --verbose                 Enable verbose mode.
+  --yaml                    Display output as yaml.
+  --json                    Display output as json.
+  --pretty                  Pretty print output  [default: True]
   --help                    Show this message and exit.
 ```
 
@@ -198,13 +198,13 @@ Usage: substra add algo [OPTIONS] PATH
   - permissions: define asset access permissions
 
 Options:
-  --yaml          Display output as yaml.
-  --json          Display output as json.
-  --pretty        Pretty print output  [default: True]
   --config PATH   Config path (default ~/.substra).
   --profile TEXT  Profile name to use.
   --user FILE     User file path to use (default ~/.substra-user).
   --verbose       Enable verbose mode.
+  --yaml          Display output as yaml.
+  --json          Display output as json.
+  --pretty        Pretty print output  [default: True]
   --help          Show this message and exit.
 ```
 
@@ -254,13 +254,13 @@ Usage: substra add compute_plan [OPTIONS] TUPLES_PATH
 
 Options:
   --objective-key TEXT  [required]
-  --yaml                Display output as yaml.
-  --json                Display output as json.
-  --pretty              Pretty print output  [default: True]
   --config PATH         Config path (default ~/.substra).
   --profile TEXT        Profile name to use.
   --user FILE           User file path to use (default ~/.substra-user).
   --verbose             Enable verbose mode.
+  --yaml                Display output as yaml.
+  --json                Display output as json.
+  --pretty              Pretty print output  [default: True]
   --help                Show this message and exit.
 ```
 
@@ -291,13 +291,13 @@ Usage: substra add aggregate_algo [OPTIONS] PATH
   - permissions: define asset access permissions
 
 Options:
-  --yaml          Display output as yaml.
-  --json          Display output as json.
-  --pretty        Pretty print output  [default: True]
   --config PATH   Config path (default ~/.substra).
   --profile TEXT  Profile name to use.
   --user FILE     User file path to use (default ~/.substra-user).
   --verbose       Enable verbose mode.
+  --yaml          Display output as yaml.
+  --json          Display output as json.
+  --pretty        Pretty print output  [default: True]
   --help          Show this message and exit.
 ```
 
@@ -328,13 +328,13 @@ Usage: substra add composite_algo [OPTIONS] PATH
   - permissions: define asset access permissions
 
 Options:
-  --yaml          Display output as yaml.
-  --json          Display output as json.
-  --pretty        Pretty print output  [default: True]
   --config PATH   Config path (default ~/.substra).
   --profile TEXT  Profile name to use.
   --user FILE     User file path to use (default ~/.substra-user).
   --verbose       Enable verbose mode.
+  --yaml          Display output as yaml.
+  --json          Display output as json.
+  --pretty        Pretty print output  [default: True]
   --help          Show this message and exit.
 ```
 
@@ -362,13 +362,13 @@ Options:
   --data-samples-path FILE  [required]
   --in-model-key TEXT       In model traintuple key.
   --tag TEXT
-  --yaml                    Display output as yaml.
-  --json                    Display output as json.
-  --pretty                  Pretty print output  [default: True]
   --config PATH             Config path (default ~/.substra).
   --profile TEXT            Profile name to use.
   --user FILE               User file path to use (default ~/.substra-user).
   --verbose                 Enable verbose mode.
+  --yaml                    Display output as yaml.
+  --json                    Display output as json.
+  --pretty                  Pretty print output  [default: True]
   --help                    Show this message and exit.
 ```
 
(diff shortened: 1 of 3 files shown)


jmorel:
I think the purpose of this TODO note was not just to group option declarations into a single decorator but also to not passe the options' values to the command at all, instead putting them into an object that is attached to the context injected into each command. The instantiation of the client and the configuration of the printers would then only depend on this object instead of multiple individual parameters.

I think it could be achieved using your common decorator and callback for each option storing the result in a new GlobalOptions object stored in ctx.

substra/cli/interface.py
```diff
@@ -148,6 +144,14 @@ def click_option_verbose(f):
     )(f)
 
 
+def click_option_global(f):
```
Kelvin-M: Maybe it is more comprehensible to call them `click_option_common` no ?

maeldebon: Yes, that's better this way, I'll rename it

substra/cli/interface.py
```diff
@@ -74,14 +74,26 @@ def display(res):
...
 
 
-# TODO profile, config, json and verbose options should be handled in a single
-#      decorator to populate a GlobalOption object stored in the context
+class GlobalConf:
+    profile = None
+    config = None
+    user = None
+    verbose = None
+    expand = None
```
jmorel: You're missing `output_format`.

substra/cli/interface.py
```diff
@@ -27,20 +27,20 @@
 from substra.sdk import user as usr
 
 
-def get_client(config_path, profile_name, user_path):
+def get_client(params):
```
jmorel: Why call it params here instead of global_conf as you did everywhere else?

tests/test_printers.py
```diff
@@ -17,6 +17,14 @@
 from substra.cli import printers
 
 
+class GlobalConf:
+    profile = None
+    config = None
+    user = None
+    verbose = None
+    expand = None
```
jmorel: Instead of re-defining this object, you could  just import it from interface.py.

substra/cli/interface.py
maeldebon: It is not required for all methods where we use click_option_common. For example, it is not used in add_data_sample, describe or download, that's why it isn't in the GlobalConf class

substra/cli/interface.py
maeldebon: Yes, I will rename it! :)

tests/test_printers.py
maeldebon: That's true, I'll do that!

substra/cli/interface.py
jmorel: The attributes we're listing here are not really useful : since we use getattr/setattr, there's no need for them. Their purpose is more to inform the developer of which attributes may be available. And for that purpose, I think that the `output_format` should be added.

maeldebon: Ok, I'll add it!

samlesu: 👍, it's better to be explicit.

I'm unresolving this comment as `output_format` is missing.

substra/cli/interface.py
```diff
@@ -74,14 +74,27 @@ def display(res):
...
+    output_format = None
+
+
+def save_to_global_conf(ctx, param, value):
+    if not hasattr(ctx, 'global_conf'):
+        setattr(ctx, 'global_conf', GlobalConf())
+    setattr(ctx.global_conf, param.name, value)
 
 
 def click_option_profile(f):
```
samlesu: I think we should renamed these functions. What do you think about `click_global_conf_profile`?
It will be easier for developers when reading the code to understand that these options are available through the `ctx.GlobalConf` instance and not as input variable.

substra/cli/printers.py
```diff
@@ -490,21 +490,21 @@ def print(self, leaderboard, expand):
 }
 
 
-def get_asset_printer(asset, output_format):
-    if output_format == 'pretty' and asset in PRINTERS:
+def get_asset_printer(asset, params):
```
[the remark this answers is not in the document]
maeldebon: So I can just input ctx.global_conf.output_format instead of ctx.global_conf?

substra/cli/interface.py
maeldebon: Yes, that's a better name I think, I'll rename the function :+1:

substra/cli/interface.py
```diff
@@ -143,11 +168,21 @@ def click_option_verbose(f):
...
     return click.option(
         '--verbose',
+        expose_value=False,
+        callback=save_to_global_conf,
         is_flag=True,
         help='Enable verbose mode.'
     )(f)
 
 
+def click_option_common(f):
```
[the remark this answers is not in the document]
maeldebon: I'll rename it yes!

I'm not sure about adding `output_format`, because it means that we will add this option even where it is not useful (for example in `describe` or `download`) as we did for expand. 🤔 
So removing them from the `GlobalConf` looks better to me, what do you think about that?

substra/cli/interface.py
```diff
@@ -74,14 +74,27 @@ def display(res):
...
+    config = None
+    user = None
+    verbose = None
+    expand = None
+    output_format = None
+
+
+def save_to_global_conf(ctx, param, value):
+    if not hasattr(ctx, 'global_conf'):
+        setattr(ctx, 'global_conf', GlobalConf())
```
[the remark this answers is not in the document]
maeldebon: I think so, but I'm not sure about how to do that :/

tests/test_cli.py
```diff
@@ -324,24 +324,6 @@ def test_command_update_data_sample(workdir, mocker):
     assert re.search(r'File ".*" does not exist\.', res)
 
 
-@pytest.mark.parametrize('params,output', [
-    ([], 'pretty\n'),
-    (['--pretty'], 'pretty\n'),
-    (['--json'], 'json\n'),
-    (['--yaml'], 'yaml\n'),
-])
-def test_option_output_format(params, output):
```
maeldebon: I removed this test because it was only testing the click.option method

references/cli.md
```diff
@@ -523,6 +526,9 @@ Options:
   --profile TEXT  Profile name to use.
   --user FILE     User file path to use (default ~/.substra-user).
   --verbose       Enable verbose mode.
+  --yaml          Display output as yaml.
+  --json          Display output as json.
+  --pretty        Pretty print output  [default: True]
```
jmorel: These options are irrelevant for describe.

references/cli.md
```diff
@@ -545,6 +551,9 @@ Options:
...
```
jmorel: As for describe, these options don't make sense for download.

references/cli.md
maeldebon: You're right, but I have to remove the output_format option from the `click_global` method to remove it from the `--help`, should we really do that? 🤔

jmorel: We could have 2 decorators, one that includes the output format and one that doesn't.

maeldebon: Yes, it could be cleaner this way, I'll work on it!

substra/cli/interface.py
```diff
@@ -193,7 +230,7 @@ def cli(ctx):
     For help using this tool, please open an issue on the Github repository:
     https://github.com/SubstraFoundation/substra-cli
     """
-    pass
+    ctx.obj = GlobalConf()
```
jmorel: I'm always wary of very generic names, I always prefer to have an explicit one. I understand that using `global_conf` may be a bit long, so what do you think of using `conf` instead?

tests/test_printers.py
```diff
@@ -15,6 +15,7 @@
 import pytest
 
 from substra.cli import printers
+from substra.cli.interface import GlobalConf
```
jmorel: I don't think these changes are needed anymore. Your last refactor removed changes from `printers.py` so the associated unit tests shouldn't change.

substra/cli/interface.py
maeldebon: Unfortunately, we cannot use another attribute's name, it is specified in the documentation that we must use `obj` to store these data. I tried to, but it doesn't worked :c 
`obj – an arbitrary object of user data.`

Link: https://click.palletsprojects.com/en/7.x/api/#click.Context

tests/test_printers.py
maeldebon: You're right, I'll remove this! 🙂

substra/cli/interface.py
jmorel: Oh thank you, I wasn't aware of this default `obj` option. Let's use this then, this is clearly the way things should be done.

tests/test_printers.py
```diff
@@ -42,7 +42,8 @@ def test_find_dict_composite_key_value_fails():
     ('foo', 'yaml', printers.YamlPrinter),
 ])
 def test_get_asset_printer(asset, output_format, printer_cls):
-    assert isinstance(printers.get_asset_printer(asset, output_format), printer_cls)
+    def get_asset_printer(ctx):  # need to use this method to pass the ctx parameter
+        assert isinstance(printers.get_asset_printer(asset, ctx.obj.output_format), printer_cls)
```
jmorel: There is an issue here: you're defining a function that will do the assert but never trigger the function. So the test passes but only because there is actual assertion.

That being said, since `get_asset_printer` isn't modified at all, you should be able to just remove all changes from this file.

maeldebon: It works, indeed. I change that in the next commit! 👍

substra/cli/interface.py
```diff
@@ -139,12 +162,26 @@ def click_option_output_format(f):
...
+    f = click_global_conf_verbose(f)
+    f = click_global_conf_user(f)
+    f = click_global_conf_profile(f)
+    f = click_global_conf_config(f)
+    return f
+
+
+def click_global_conf_with_output_format(f):
+    f = click_global_conf_output_format(f)
+    click_global_conf(f)
```
jmorel: Isn't `f =` missing here?

maeldebon: Yes, I forgot to add it 🙁

substra/cli/interface.py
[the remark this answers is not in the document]
samlesu: Has been discussed directly.

substra/cli/printers.py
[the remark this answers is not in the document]
samlesu: Sorry for the late response, yes.

Issue #19, opened 2019-11-05
SDK: when adding an asset, don't always add permissions field


samlesu:
Currently when adding an asset a `permissions` is added by default. This should not be the case for the traintuple and testtuple assets for instance.

Adding a `permissions` field should be done explicitly when it's required.

maeldebon:
Done in #49

PR #49, opened 2019-12-05
Adding asset without permissions field - Issue #19

maeldebon:
Currently, when an asset is added, a `permissions` field is automatically created, even if the asset doesn't needs one (for example, `traintuple` or `testtuple`).
This PR aims to avoid adding a default permissions field for these assets.

Issue related to this PR: #19

### DIFF
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -40,6 +40,20 @@ def _flatten_permissions(data, field_name):
     return data
 
 
+def _update_permissions_field(data, permissions_field='permissions'):
+    if permissions_field not in data:
+        default_permissions = {
+            'public': False,
+            'authorized_ids': [],
+        }
+        data[permissions_field] = default_permissions
+    # XXX workaround because backend accepts only Form Data body. This is due to
+    #     the fact that backend expects both file objects and payload in the
+    #     same request
+    data = _flatten_permissions(data, field_name=permissions_field)
+    return data
+
+
 class Client(object):
 
     def __init__(self, config_path=None, profile_name=None, user_path=None):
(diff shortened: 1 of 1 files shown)


substra/sdk/client.py
```diff
@@ -461,9 +467,7 @@ def add_composite_traintuple(self, data, timeout=False, exist_ok=False):
         If `exist_ok` is true, `AlreadyExists` exceptions will be ignored and the
         existing asset will be returned.
         """
-        res = self._add(assets.COMPOSITE_TRAINTUPLE, data, timeout=timeout,
-                        exist_ok=exist_ok,
-                        permissions_field='out_model_trunk_permissions')
+        res = self._add(assets.COMPOSITE_TRAINTUPLE, data, timeout=timeout, exist_ok=exist_ok)
```
jmorel: There is a permission field in this case: `out_model_trunk_permissions`. It should also be populated with a default value if not present.

substra/sdk/client.py
```diff
@@ -110,6 +110,20 @@ def set_user(self):
                 })
                 self.client.set_config(self._current_profile, self._profile_name)
 
+    def update_permissions_field(self, data, permissions_field='permissions'):
+        if permissions_field:
```
jmorel: The purpose of this function is to provide a default value for a permissions field. It shouldn't have to handle the case in which there is no permission field at all. It is the job of the parent function to decide whether to call this function or not.

substra/sdk/client.py
```diff
@@ -110,6 +110,20 @@ def set_user(self):
                 })
                 self.client.set_config(self._current_profile, self._profile_name)
 
+    def update_permissions_field(self, data, permissions_field='permissions'):
```
samlesu: This function has no real reason to be part of the Client class: it does not interact with the server, it modifies a dict object. So it could be moved outside of the client class.

This function must be private: API users should not be aware of this function.

samlesu:
> I left a couple of remarks on this PR regarding your chosen implementation, but I also wanted to point out that there was another solution that entailed fewer changes.

That's my fault, I lead Maël in that direction. More changes indeed, but specific cases will be handled at a higher level and not in the "generic" function. But that's pretty minor, we don't have to do that now.

substra/sdk/client.py
maeldebon: I'm agree!

Issue #50, opened 2019-12-05
KeyError with full traceback


maeldebon:
When I tried to add a traintuple without the "keys" field in the --data-samples-path field, I got a full traceback instead of only an error message:

```
Traceback (most recent call last):
  File "/Users/mael/.venv/substra/bin/substra", line 11, in <module>
    load_entry_point('substra', 'console_scripts', 'substra')()
  File "/Users/mael/.venv/substra/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/mael/.venv/substra/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/mael/.venv/substra/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/mael/.venv/substra/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/mael/.venv/substra/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/mael/.venv/substra/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/mael/.venv/substra/lib/python3.7/site-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/Users/mael/substra/substra/substra/cli/interface.py", line 172, in wrapper
    return fn(*args, **kwargs)
  File "/Users/mael/substra/substra/substra/cli/interface.py", line 611, in add_traintuple
    data['train_data_sample_keys'] = data_samples['keys']
KeyError: 'keys'
```

jmorel:
Fixed by #53

PR #53, opened 2019-12-05
Avoid KeyError when keys are missing in the data_samples file

maeldebon:
Currently, when the `keys` attribute is missing in the data_samples file we try to use, an error is raised and a full traceback is returned.
The commands impacted are the followings:
- add_objective
- add_traintuple
- add_composite_traintuple
- add_testtuple
- update_data_sample

🔗 Related issue: #50

### DIFF
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -416,7 +416,7 @@ def add_objective(ctx, data, dataset_key, data_samples):
         data['test_data_manager_key'] = dataset_key
 
     if data_samples:
-        data['test_data_sample_keys'] = data_samples['keys']
+        data['test_data_sample_keys'] = data_samples.get('keys', None)
 
     res = client.add_objective(data)
     printer = printers.get_asset_printer(assets.OBJECTIVE, ctx.obj.output_format)
@@ -626,7 +626,7 @@ def add_traintuple(ctx, objective_key, algo_key, dataset_key, data_samples, in_m
     }
 
     if data_samples:
-        data['train_data_sample_keys'] = data_samples['keys']
+        data['train_data_sample_keys'] = data_samples.get('keys', None)
 
     if tag:
         data['tag'] = tag
@@ -734,7 +734,7 @@ def add_composite_traintuple(ctx, objective_key, algo_key, dataset_key, data_sam
     }
 
     if data_samples:
-        data['train_data_sample_keys'] = data_samples['keys']
+        data['train_data_sample_keys'] = data_samples.get('keys', None)
 
     if out_trunk_model_permissions:
         data['out_trunk_model_permissions'] = out_trunk_model_permissions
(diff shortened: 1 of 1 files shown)


substra/cli/interface.py
```diff
@@ -1068,7 +1076,7 @@ def update_data_sample(ctx, data_samples, dataset_key):
     - keys: list of data sample keys
     """
     client = get_client(ctx.obj)
-    res = client.link_dataset_with_data_samples(dataset_key, data_samples['keys'])
+    res = client.link_dataset_with_data_samples(dataset_key, load_data_samples_keys(data_samples))
```
jmorel: There is no "--data-samples-path" option for this command, the error message raises by `load_data_samples_keys` would look very wrong.

maeldebon: Indeed, but there is an argument called 'data_samples', so we're supposed to get this object, because it is required

jmorel: Yes absolutely, we just need the proper error message.

jmorel:
I think there's still something missing. Here is the error I get when I give the wrong file type:
```
substra update data_sample README.md
Usage: substra update data_sample [OPTIONS] DATA_SAMPLES_PATH

Error: Invalid value for "DATA_SAMPLES_PATH": File "README.md" is not a valid JSON file.
```

And when I give the wrong content:
```
substra update data_sample compute_plan_tuples.json --dataset-key=toto
Usage: substra update data_sample [OPTIONS] DATA_SAMPLES_PATH

Error: Invalid value for data_samples: The file specified must contain a 'keys' attribute
```

I'd expect the error message to start with the same `Error: Invalid value for "DATA_SAMPLES_PATH":` prefix and also end with a `.`.

Same thing with the option: I get 
````
Error: Invalid value for "--data-samples-path": File "/Users/jeremy/substra/substra/README.md" is not a valid JSON file.
```
instead of
```
Error: Invalid value for --data-samples-path: The file specified must contain a 'keys' attribute
```

Also, in order to keep in line with the tone of error messages, we could simplify the message to `File must contain a 'keys' attribute.` (no article, no 'specified')

maeldebon:
> I think there's still something missing. Here is the error I get when I give the wrong file type:
> 
> ```
> substra update data_sample README.md
> Usage: substra update data_sample [OPTIONS] DATA_SAMPLES_PATH
> 
> Error: Invalid value for "DATA_SAMPLES_PATH": File "README.md" is not a valid JSON file.
> ```
> 
> And when I give the wrong content:
> 
> ```
> substra update data_sample compute_plan_tuples.json --dataset-key=toto
> Usage: substra update data_sample [OPTIONS] DATA_SAMPLES_PATH
> 
> Error: Invalid value for data_samples: The file specified must contain a 'keys' attribute
> ```
> 
> I'd expect the error message to start with the same `Error: Invalid value for "DATA_SAMPLES_PATH":` prefix and also end with a `.`.
> 
> Same thing with the option: I get
> 
> ```
> Error: Invalid value for "--data-samples-path": File "/Users/jeremy/substra/substra/README.md" is not a valid JSON file.
> ```
> 
> instead of
> 
> ```
> Error: Invalid value for --data-samples-path: The file specified must contain a 'keys' attribute
> ```
> 
> Also, in order to keep in line with the tone of error messages, we could simplify the message to `File must contain a 'keys' attribute.` (no article, no 'specified')

Okay, I'll do that! 🙂

Issue #55, opened 2019-12-06
New cancel compute plan command


jmorel:
In link with the "fail fast" mission, we need to have in both the cli and sdk cancel compute plan commands / methods with the following signatures:

In the cli:
```
$ substra cancel compute_plan <compute_plan_id>
```
This command should support the same options as get compute plan.

In the sdk:
```python
client.cancel_compute_plan(compute_plan_id)
```

Compute plans will get a new "status" property. It will have to be added to the printers in both list and single item displays.

And of course, all of this will need to be tested properly.

Kelvin-M:
@jmorel This PR (https://github.com/SubstraFoundation/substra/pull/56) was merged, did it fix your issue ?

jmorel:
Absolutely! I'm closing the issue.

PR #56, opened 2019-12-06
New cancel compute_plan command

maeldebon:
In link with the "fail fast" mission, we need to have in both the CLI and SDK cancel compute plan commands / methods.

🔗 Related issue: #55 

Companion PR:
- [tests](https://github.com/SubstraFoundation/substra-tests/pull/26)
- [backend](https://github.com/SubstraFoundation/substra-backend/pull/65)
- [chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/53)
- [frontend](https://github.com/SubstraFoundation/substra-frontend/pull/33)

### DIFF
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -1046,7 +1046,7 @@ def cancel_compute_plan(ctx, compute_plan_id):
     client = get_client(ctx.obj)
     # method must exist in sdk
     res = client.cancel_compute_plan(compute_plan_id)
-    printer = printers.get_asset_printer(compute_plan_id, ctx.obj.output_format)
+    printer = printers.get_asset_printer(assets.COMPUTE_PLAN, ctx.obj.output_format)
     printer.print(res, profile=ctx.obj.profile)
 
 
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -1032,6 +1032,26 @@ def run_local(algo, train_opener, test_opener, metrics, rank,
 
 @cli.group()
 @click.pass_context
+def cancel(ctx):
+    """Cancel execution of an asset."""
+    pass
+
+
+@cancel.command('compute_plan')
+@click.argument('compute_plan_id', type=click.STRING)
+@click_global_conf_with_output_format
+@click.pass_context
+def cancel_compute_plan(ctx, compute_plan_id):
+    """Cancel execution of a compute plan."""
+    client = get_client(ctx.obj)
+    # method must exist in sdk
+    res = client.cancel_compute_plan(compute_plan_id)
+    printer = printers.get_asset_printer(compute_plan_id, ctx.obj.output_format)
+    printer.print(res, profile=ctx.obj.profile)
+
+
+@cli.group()
+@click.pass_context
 def update(ctx):
     """Update asset."""
     pass
--- a/substra/cli/printers.py
+++ b/substra/cli/printers.py
@@ -248,7 +248,7 @@ class ComputePlanPrinter(AssetPrinter):
         KeysField('Composite traintuple keys', 'compositeTraintupleKeys'),
         KeysField('Aggregatetuple keys', 'aggregatetupleKeys'),
         KeysField('Testtuple keys', 'testtupleKeys'),
-        KeysField('Status', 'status'),
+        Field('Status', 'status'),
     )
 
     def print_messages(self, item, profile=None):
--- a/substra/cli/printers.py
+++ b/substra/cli/printers.py
@@ -241,12 +241,14 @@ class ComputePlanPrinter(AssetPrinter):
         CountField('Composite traintuples count', 'compositeTraintupleKeys'),
         CountField('Aggregatetuples count', 'aggregatetupleKeys'),
         CountField('Testtuples count', 'testtupleKeys'),
+        Field('Status', 'status'),
     )
     single_fields = (
         KeysField('Traintuple keys', 'traintupleKeys'),
         KeysField('Composite traintuple keys', 'compositeTraintupleKeys'),
         KeysField('Aggregatetuple keys', 'aggregatetupleKeys'),
         KeysField('Testtuple keys', 'testtupleKeys'),
+        KeysField('Status', 'status'),
     )
 
     def print_messages(self, item, profile=None):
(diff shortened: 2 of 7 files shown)


substra/cli/interface.py
```diff
@@ -1038,6 +1038,27 @@ def run_local(algo, train_opener, test_opener, metrics, rank,
...
+@cli.group()
+@click.pass_context
+def cancel(ctx):
+    """Cancel execution of an asset."""
+    pass
+
+
+@cancel.command('compute_plan')
+@click.argument('compute_plan_id', type=click.STRING)
+@click_global_conf
```
samlesu: it looks the output format option is missing?

substra/cli/interface.py
```diff
@@ -1038,6 +1038,27 @@ def run_local(algo, train_opener, test_opener, metrics, rank,
...
+@click.pass_context
+def cancel_compute_plan(ctx, compute_plan_id):
+    """Cancel execution of a compute plan."""
+    client = get_client(ctx.obj)
+    # method must exist in sdk
+    method = getattr(client, f'cancel_compute_plan')
+    res = method(compute_plan_id)
```
samlesu: Why are you using a `getattr` method here? It seems not needed.

substra/cli/interface.py
```diff
@@ -1038,6 +1038,27 @@ def run_local(algo, train_opener, test_opener, metrics, rank,
...
+    printer = printers.get_asset_printer(compute_plan_id, ctx.obj.output_format)
+    printer.print(res, profile=ctx.obj.profile)
```
samlesu: I'm not 100% sure regarding the output of the cancel compute plan command. To be checked with the person in charge of the chaincode implemention or in the mission specification.

substra/sdk/client.py
```diff
@@ -764,3 +764,7 @@ def leaderboard(self, objective_key, sort='desc'):
         """Get objective leaderboard"""
         return self.client.request('get', assets.OBJECTIVE, f'{objective_key}/leaderboard',
                                    params={'sort': sort})
+
+    def cancel_compute_plan(self, compute_plan_id):
+        """Cancel execution of compute plan."""
+        return self.client.get(assets.COMPUTE_PLAN, compute_plan_id)
```
samlesu: This will get a compute plan and not cancel an existing one. 
A new route will be created in the backend to handle the compute plan execution. The route will be something like `computeplan/<id>/cancel`.

substra/cli/interface.py
maeldebon: You're right, I forgot to add it here

substra/cli/interface.py
maeldebon: Okay, perfect so! 🙂

substra/sdk/client.py
maeldebon: Okay, but I'm not sure I can change that until this is done in the backend?

substra/cli/interface.py
maeldebon: You're absolutely right, I'll remove it

substra/sdk/client.py
samlesu: It has been indeed not defined in the specification, the PR for the backend is now available:
https://github.com/SubstraFoundation/substra-backend/pull/65
It may change in the future after the review but you can start with: `POST /computeplan/<id>/cancel`

tests/sdk/test_cancel.py
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .. import datastore
+from .utils import mock_requests
+
+
+def test_cancel_compute_plan(client, mocker):
+    item = getattr(datastore, 'COMPUTE_PLAN')
+    method = getattr(client, f'cancel_compute_plan')
```
samlesu: Why did you use a `getattr` method here? I don't think it's needed.

maeldebon: Yes, I removed them

tests/sdk/test_cancel.py
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .. import datastore
+from .utils import mock_requests
+
+
+def test_cancel_compute_plan(client, mocker):
+    item = datastore.COMPUTE_PLAN
+    method = client.cancel_compute_plan
```
samlesu: Very minor: I think there is no need to create variable for these 2 objects.

maeldebon: Indeed, that's cleaner this way 👍

jmorel:
@maeldebon could you add the 'status' field to the compute plan printer? For both list and single modes.

substra/cli/interface.py
```diff
@@ -1030,6 +1030,26 @@ def run_local(algo, train_opener, test_opener, metrics, rank,
...
+@click.argument('compute_plan_id', type=click.STRING)
+@click_global_conf_with_output_format
+@click.pass_context
+def cancel_compute_plan(ctx, compute_plan_id):
+    """Cancel execution of a compute plan."""
+    client = get_client(ctx.obj)
+    # method must exist in sdk
+    res = client.cancel_compute_plan(compute_plan_id)
+    printer = printers.get_asset_printer(compute_plan_id, ctx.obj.output_format)
+    printer.print(res, profile=ctx.obj.profile)
```
jmorel: I think you're missing the output_format option here, as a result it always prints the response as JSON.

substra/cli/printers.py
```diff
@@ -241,12 +241,14 @@ class ComputePlanPrinter(AssetPrinter):
...
         CountField('Aggregatetuples count', 'aggregatetupleKeys'),
         CountField('Testtuples count', 'testtupleKeys'),
+        Field('Status', 'status'),
     )
     single_fields = (
         KeysField('Traintuple keys', 'traintupleKeys'),
         KeysField('Composite traintuple keys', 'compositeTraintupleKeys'),
         KeysField('Aggregatetuple keys', 'aggregatetupleKeys'),
         KeysField('Testtuple keys', 'testtupleKeys'),
+        KeysField('Status', 'status'),
```
jmorel: This isn't a keysfield but rather a Field. Otherwise we'll get an output like:
```
STATUS                          5 keys
```

jmorel:
@maeldebon it looks like the tests aren't passing, could you have a look at it?

maeldebon:
> @maeldebon it looks like the tests aren't passing, could you have a look at it?

Yes, I'm currently trying to fix that 🙂

PR #57, opened 2019-12-06
Add missing tests

maeldebon:
For now, some commands are not tested at all.
This PR aims to add the missing tests.

### DIFF
--- a/tests/sdk/test_add.py
+++ b/tests/sdk/test_add.py
@@ -11,8 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import re
-
 import pytest
 import substra
 
--- a/tests/sdk/test_add.py
+++ b/tests/sdk/test_add.py
@@ -109,13 +109,10 @@ def test_add_data_sample_already_exists(client, data_sample_query, mocker):
     m.assert_called()
 
 
-def test_add_data_sample_with_paths(client, data_samples_query, mocker):
-    server_response = [{"key": "42"}]
-    m = mock_requests(mocker, "post", response=server_response)
+# We try to add multiple data samples instead of a single one
+def test_add_data_sample_with_paths(client, data_samples_query):
     with pytest.raises(ValueError):
         client.add_data_sample(data_samples_query)
-
-    assert m.is_called()
 
 
 def test_add_data_samples(client, data_samples_query, mocker):
@@ -127,10 +124,7 @@ def test_add_data_samples(client, data_samples_query, mocker):
     m.assert_called()
 
 
-def test_add_data_samples_with_path(client, data_sample_query, mocker):
-    server_response = [{"key": "42"}]
-    m = mock_requests(mocker, "post", response=server_response)
+# We try to add a single data sample instead of multiple ones
+def test_add_data_samples_with_path(client, data_sample_query):
     with pytest.raises(ValueError):
         client.add_data_samples(data_sample_query)
-
-    assert m.is_called()
--- a/tests/sdk/test_add.py
+++ b/tests/sdk/test_add.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import re
 
 import pytest
 import substra
@@ -110,6 +111,15 @@ def test_add_data_sample_already_exists(client, data_sample_query, mocker):
     m.assert_called()
 
 
+def test_add_data_sample_with_paths(client, data_samples_query, mocker):
+    server_response = [{"key": "42"}]
+    m = mock_requests(mocker, "post", response=server_response)
+    with pytest.raises(ValueError):
+        client.add_data_sample(data_samples_query)
+
+    assert m.is_called()
+
+
 def test_add_data_samples(client, data_samples_query, mocker):
     server_response = [{"key": "42"}]
     m = mock_requests(mocker, "post", response=server_response)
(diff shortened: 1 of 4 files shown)


tests/test_cli.py
```diff
@@ -207,6 +213,18 @@ def test_command_add_composite_traintuple(mocker, workdir, params, message, exit
...
+def test_command_add_aggregatetuple(mocker, workdir):
+    m = mock_client_call(mocker, 'add_aggregatetuple', response={})
+    client_execute(workdir, ['add', 'aggregatetuple', '--objective-key', 'foo', '--algo-key', 'foo',
+                             '--in-model-key', 'foo', '--worker', 'foo'])
+    assert m.is_called()
+
+    m = mock_client_call(mocker, 'add_aggregatetuple', response={})
+    client_execute(workdir, ['add', 'aggregatetuple', '--objective-key', 'foo', '--algo-key', 'foo',
+                             '--in-model-key', 'foo', '--in-model-key', 'bar', '--worker', 'foo'])
+    assert m.is_called()
```
jmorel: You could use parametrize here.

maeldebon: Yes, I'll add this!

tests/sdk/test_add.py
```diff
@@ -93,27 +94,37 @@ def test_add_composite_algo(client, algo_query, mocker):
     assert m_get.is_called()
 
 
-def test_add_data_sample(client, data_sample_query, mocker):
+def test_add_data_sample(client, data_sample_query, data_samples_query, mocker):
```
samlesu: When writing unit tests we should try to follow the rule: a test should check one thing or test one feature/a single case. 

Tests are then easier to read (we can quickly understand what's the goal of test) and it's then easier to understand what is the issue when a test is broken.

So you should not extand or group existing tests but create new ones.

maeldebon: Maybe I can check the existing tests and apply this rule?

samlesu: That could be done later but it's not a priority right now and it's better to have a small scope in each PR. So in my opinion, this PR should only add a few missing tests and not change the existing ones.

maeldebon: Okay!

jmorel:
@maeldebon it looks like this branch includes commits from #61 and master. Could you clean it up?

maeldebon:
> @maeldebon it looks like this branch includes commits from #61 and master. Could you clean it up?

Absolutely not sure about what I've done 😕

tests/sdk/test_add.py
```diff
@@ -110,10 +109,28 @@ def test_add_data_sample_already_exists(client, data_sample_query, mocker):
     assert m.is_called()
 
 
+def test_add_data_sample_with_paths(client, data_samples_query, mocker):
+    server_response = [{"key": "42"}]
+    m = mock_requests(mocker, "post", response=server_response)
+    with pytest.raises(ValueError):
+        client.add_data_sample(data_samples_query)
+
+    assert m.is_called()
```
jmorel: It took me a while to understand what this was doing. Could you add a comment explaining that you're intentionally passing `data_samples_query` instead of `data_sample_query` in order to check that you get an exception?

Also, I don't think `m` should be called at all. If it is then something went wrong in the validation.

tests/sdk/test_add.py
```diff
@@ -110,10 +109,28 @@ def test_add_data_sample_already_exists(client, data_sample_query, mocker):
...
     assert m.is_called()
+
+
+def test_add_data_samples_with_path(client, data_sample_query, mocker):
+    server_response = [{"key": "42"}]
+    m = mock_requests(mocker, "post", response=server_response)
+    with pytest.raises(ValueError):
+        client.add_data_samples(data_sample_query)
+
+    assert m.is_called()
```
jmorel: Almost same comment as above: could you add a comment explaining what you're testing exactly and could you double check that m is called?

tests/sdk/test_add.py
maeldebon: I'll add comments, it will indeed be clearer.
But I'm not sure about the second part of your comment, should I just remove the variable (and the related assert) or only be sure that it is properly called and checked?

jmorel: According to the code of  `client.add_data_sample`, the validation raising the ValueError occurs before any call is made to the API. Which means that the mocker should effectively never be called at all.

I tried this test:

```python
def test_add_data_sample_with_paths(client, data_samples_query, mocker):
    server_response = [{"key": "42"}]
    m = mock_requests(mocker, "post", response=server_response)
    assert m.is_called()
```

And it passes ! Even though there is absolutely no call!
This is a big issue, I'll create a new issue for it.

That being said, in this very case, since we're expecting the validation to break the call to add_data_sample, we don't need a mock response at all, the test could be simplified as:
```python
def test_add_data_sample_with_paths(client, data_samples_query, mocker):
    with pytest.raises(ValueError):
        client.add_data_sample(data_samples_query)
```
And all tests also expecting the validation to raise an exception could be simplified in a similar way.

maeldebon: Indeed, I'll remove the `m.is_called()` and the `mock_requests` for these two tests!
I'll have a look to the others test that raises exceptions as well 🙂

tests/sdk/test_list.py
```diff
@@ -65,11 +65,10 @@ def test_list_asset_with_filters(client, mocker):
 
 def test_list_asset_with_filters_failure(client, mocker):
     items = [datastore.ALGO]
-    m = mock_requests(mocker, "get", response=[items])
+    mock_requests(mocker, "get", response=[items])
```
jmorel: It looks like it's not necessary at all to mock requests.get here, the ValueError raised by list_algo seems to occur before the call to requests.get is made.

You could probably remove the line entirely.

Issue #58, opened 2019-12-09
Add debug for the CLI's tests


maeldebon:
Currently, this is really hard to debug the CLI.
It could be good to add some information when a test crashes.
For instance, the full traceback, the exc_code, etc.

Kelvin-M:
You can also use pytest option with `addopts` option like this  `python setup.py test --addopts='--capture no'`. 

You can also launch one test like this : 
`python setup.py test --addopts='tests/sdk/test_rest_client.py::test_add_exist_ok' `

samlesu:
Fixed by #61

PR #61, opened 2019-12-09
Add debug for CLI's tests

maeldebon:
As specified in the related issue, this is sometimes hard to debug the CLI.
It could be good to add some information when a test crashes.
For instance, the full traceback, the exc_code, etc.

🔗 Related issue: #58

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -15,6 +15,7 @@
 import json
 import re
 from unittest import mock
+import traceback
 
 import click
 from click.testing import CliRunner
@@ -36,15 +37,12 @@ def workdir(tmp_path):
 def execute(command, exit_code=0):
     runner = CliRunner()
     result = runner.invoke(cli, command)
-    if result.exception:
-        import traceback
-        print('# Command: ', command)
-        print('# Exception: ', result.exception)
-        print('# Exit code: ', result.exit_code)
+    debugging = result.output
+    if exit_code == 0 and result.exit_code != 0 and result.exc_info:
         _, _, tb = result.exc_info
-        print('# Traceback: ', traceback.print_tb(tb))
-    assert result.exit_code == exit_code, result.output
-    return result.output
+        debugging = traceback.print_tb(tb)
+    assert result.exit_code == exit_code, debugging
+    return debugging
 
 
 def client_execute(directory, command, exit_code=0):
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -14,6 +14,7 @@
 
 import json
 import re
+import sys
 from unittest import mock
 import traceback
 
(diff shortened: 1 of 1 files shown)


tests/test_cli.py
```diff
@@ -36,8 +37,12 @@ def workdir(tmp_path):
 def execute(command, exit_code=0):
     runner = CliRunner()
     result = runner.invoke(cli, command)
-    assert result.exit_code == exit_code, result.output
-    return result.output
+    debugging = result.output
+    if exit_code == 0 and result.exit_code != 0 and result.exc_info:
+        _, _, tb = result.exc_info
+        debugging = traceback.print_tb(tb)
```
jmorel: This will return as follows: 
```
command = ['config', 'http://foo', '--profile', 'foo', '--config', '/private/var/folders/z9/tjd43wyd11j9h9tn51m84jjm0000gn/T/pytest-of-jeremy/pytest-14/test_command_config0/substra-cli/cli.cfg', ...], exit_code = 0

    def execute(command, exit_code=0):
        runner = CliRunner()
        result = runner.invoke(cli, command)
        debugging = result.output
        if exit_code == 0 and result.exit_code != 0 and result.exc_info:
            _, _, tb = result.exc_info
            debugging = traceback.print_tb(tb)
>       assert result.exit_code == exit_code, debugging
E       AssertionError: None
E       assert 1 == 0
E         -1
E         +0

tests/test_cli.py:44: AssertionError
----------------------------------------------------------------------------------------------------------------------------- Captured stderr call -----------------------------------------------------------------------------------------------------------------------------
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/testing.py", line 326, in invoke
    cli.main(args=args or (), prog_name=prog_name, **extra)
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 1137, in invoke
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 956, in invoke
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 555, in invoke
  File "/Users/jeremy/substra/substra/substra/cli/interface.py", line 250, in add_profile_to_config
    raise Exception('nope')
```

You'll notice the `AssertionError: None`. This is because the print_tb doesn't return anything, it only prints the traceback. As a result, the traceback isn't next to the assertion error but in the `captured stderr call` section. If multiple test fail, all outputs will be grouped in this section, not really facilitating the debug (there is no way to know which test prompted which traceback).

Using format_tb doesn't solve the issue though, I guess in other to have a proper output you could have a look at pytest's "fail" function and pytest hooks (https://docs.pytest.org/en/latest/example/simple.html#post-process-test-reports-failures) I have no idea how to do it though, this is really about reading a lot of docs and trying out solutions.

maeldebon: I can return `print_tb` in the `stdout` call section, by using this:
`debugging = traceback.print_tb(tb, file=sys.stdout)`.
But I still have the "AssertionError: None" indeed 🤔 

I'll try to use pytest.fail so 🙂

tests/test_cli.py
```diff
@@ -37,12 +38,12 @@ def workdir(tmp_path):
...
     runner = CliRunner()
     result = runner.invoke(cli, command)
-    debugging = result.output
     if exit_code == 0 and result.exit_code != 0 and result.exc_info:
         _, _, tb = result.exc_info
-        debugging = traceback.print_tb(tb)
-    assert result.exit_code == exit_code, debugging
-    return debugging
+        traceback.print_tb(tb, file=sys.stdout)
+        pytest.fail(str(result.exception))
```
samlesu: It would be worth adding a small comment describing why this piece of code has been added.

maeldebon: Indeed, I'll add that

PR #62, opened 2019-12-09
doc: add missing rank attribute for add_traintuple method

samlesu:
This issue has been reported by @camillemarini and @jmorel.
It has been also fixed in the PR #52 that is currently under review. 

For bug fixes we should try to create independent PRs. It will allow to review them faster and then to merge them faster.


samlesu:
> Rank is also missing for add_composite_traintuple. Could you add it there as well?

Good catch thanks!

PR #64, opened 2019-12-09
Doc: add missing fields

### LINKED ISSUE #63: Doc: missing fields
- compute_plan:
  -  aggregatuple: `aggregatuple_id`
  - composite_traintuple: `composite_traintuple_id` and `permissions.public`

samlesu:
Fix #63

### DIFF
--- a/references/sdk.md
+++ b/references/sdk.md
@@ -267,6 +267,10 @@ Create new composite traintuple asset.
     "compute_plan_id": str,
 }
 ```
+
+As specified in the data dict structure, output trunk models cannot be made
+public.
+
 If `exist_ok` is true, `AlreadyExists` exceptions will be ignored and the
 existing asset will be returned.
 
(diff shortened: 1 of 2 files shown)


references/sdk.md
```diff
@@ -303,25 +303,28 @@ Data is a dict object with the following schema:
...
     }],
     "composite_traintuples": list[{
+        "composite_traintuple_id": str,
         "algo_key": str,
         "data_manager_key": str,
         "train_data_sample_keys": list[str],
         "in_head_model_id": str,
         "in_trunk_model_id": str,
         "out_trunk_model_permissions": {
+            "public": bool,
```
jmorel: The public will always be ignored, it will always be private. I think the documentation shouldn't mention a non existing parameter.

samlesu: I agree with you, it should not be there if it's ignored. But, I don't think that the public field is ignored. I looked at the chaincode implementation and it is handled like other permissions objects.

jmorel: True, but the backend always passes public set to False (`compositetraintuple.py` L45). So the user input is effectively ignored.

samlesu: Ok thanks, I will add a message then to explain this particular case.

PR #79, opened 2020-01-17
add server gateway (bad, timout, unavailable) errors support and retry

inalgnu:
This PR aims to be resilient to backend bad gateway errors, that could happen if server is restarting or down for a short time.

Having a retry mechanism will allow not stoping the current script if the server is temporarily unavailable.

### DIFF
--- a/substra/sdk/rest_client.py
+++ b/substra/sdk/rest_client.py
@@ -134,10 +134,14 @@ class Client():
             if e.response.status_code == 500:
                 raise exceptions.InternalServerError.from_request_exception(e)
 
+            if e.response.status_code in [502, 503, 504]:
+                raise exceptions.GatewayTimeout.from_request_exception(e)
+
             raise exceptions.HTTPError.from_request_exception(e)
 
         return r
 
+    @utils.retry_on_exception(exceptions=(exceptions.GatewayTimeout))
     def request(self, request_name, asset_name, path=None, json_response=True,
                 **request_kwargs):
         """Base request."""
--- a/substra/sdk/rest_client.py
+++ b/substra/sdk/rest_client.py
@@ -135,13 +135,13 @@ class Client():
                 raise exceptions.InternalServerError.from_request_exception(e)
 
             if e.response.status_code in [502, 503, 504]:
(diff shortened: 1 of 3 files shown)


substra/sdk/rest_client.py
```diff
@@ -134,10 +134,14 @@ def _request(self, request_name, url, **request_kwargs):
...
                 raise exceptions.InternalServerError.from_request_exception(e)
 
+            if e.response.status_code in [502, 503, 504]:
+                raise exceptions.GatewayTimeout.from_request_exception(e)
+
             raise exceptions.HTTPError.from_request_exception(e)
 
         return r
 
+    @utils.retry_on_exception(exceptions=(exceptions.GatewayTimeout))
```
jmorel: You may also want to add a retry on `exceptions.Timeout` that is being raised when requests raises its own `Timeout`.

samlesu: That's a good comment!

IMO we don't want to retry on Timeout: in case of timeout we don't know if the request has been processed by the server and a retry might not be the good solution.

We want to retry when the server could not receive and process the request.

Maybe the error naming is misleading? GatewayError seems better.

502: BadGateway
503: ServiceUnavailable
504: GatewayTimeout

Maybe we should exclude 504? Or have dedicated exceptions. It could help for debugging purposes.

samlesu:
@inalgnu have you experienced the 3 kind of errors (502, 503 and 504)?

inalgnu:
> @inalgnu have you experienced the 3 kind of errors (502, 503 and 504)?

@samlesu yes

substra/sdk/rest_client.py
inalgnu: as discussed, I've updated exception name here fc3473e

samlesu:
The operation "create asset" is not idempotent: creating twice the same asset will result in a conflict error.

When retrying on 502, 503, 504 errors, the server may process twice a "create asset" operation that may lead to a conflict error.

When creating an asset, ideally we should only retry requests that have not been delivered to the server.

As a short term fix, let's retry all requests on 502, 503 and 504 errors.

Issue #83, opened 2020-01-22
mock_requests returns a mock that is always called


jmorel:
In lots of places, we use mock_requests and check that the returned mock has been called. These `assert m.is_called()` are present everywhere. However it seems they don't work as we think they do. 

Here is a test that shouldn't pass:

```python
def test_foo(mocker):
    m = mock_requests(mocker, "post", response='foo')
    assert m.is_called()
```

But it does!

We need to investigate why it passes and update the mock_requests so that it works as we think it should.

maeldebon:
Great catch! 👏 
I just checked the code of `mock_requests` and, indeed, a `m` is always returned, even when an exception is raised :

```
def mock_requests_response(response=None, status=200, headers=None):
    headers = headers or {}
    m = mock.MagicMock(spec=requests.Response)
    m.status_code = status
    m.headers = headers
    m.text = str(response)
    m.json = mock.MagicMock(return_value=response, headers=headers)

    if status not in (200, 201):
        exception = requests.exceptions.HTTPError(str(status), response=m)
        m.raise_for_status = mock.MagicMock(side_effect=exception)

    return m
```

PR #85, opened 2020-01-22
Replace is_called() by assert_called() in tests

maeldebon:
🔗 Related issue: #83 

Currently, when using `is_called()` in the tests, the assert is always passing, even when it shouldn't.
As an example (provided in the issue), this test is passing but should not :
```python
def test_foo(mocker):
    m = mock_requests(mocker, "post", response='foo')
    assert m.is_called()
```

When checking the documentation of unittest.mock (https://docs.python.org/3/library/unittest.mock.html), it looks like this method doesn't exists at all.
I think we should use `assert_called()`, which is the assert we actually wanted.

### DIFF
--- a/tests/sdk/test_list.py
+++ b/tests/sdk/test_list.py
@@ -39,7 +39,7 @@ def test_list_asset(asset_name, client, mocker):
     response = method()
 
     assert response == [item]
-    assert m.is_called()
+    m.assert_called()
 
 
 def test_list_asset_flatten(client, mocker):
@@ -49,7 +49,7 @@ def test_list_asset_flatten(client, mocker):
     response = client.list_algo()
 
     assert response == items
-    assert m.is_called()
+    m.assert_called()
 
 
 def test_list_asset_with_filters(client, mocker):
@@ -60,7 +60,7 @@ def test_list_asset_with_filters(client, mocker):
     response = client.list_algo(filters)
 
     assert response == items
-    assert m.is_called()
+    m.assert_called()
 
 
 def test_list_asset_with_filters_failure(client, mocker):
@@ -71,5 +71,5 @@ def test_list_asset_with_filters_failure(client, mocker):
     with pytest.raises(ValueError) as exc_info:
         client.list_algo(filters)
 
-    assert m.is_called()
+    m.assert_not_called()
     assert str(exc_info.value).startswith("Cannot load filters")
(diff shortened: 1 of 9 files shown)


tests/test_cli.py
```diff
@@ -204,21 +204,20 @@ def test_command_add_objective(workdir, mocker):
     (['--trunk-model-key', 'e'], r'The --head-model-key option is required', 2)
 ])
 def test_command_add_composite_traintuple(mocker, workdir, params, message, exit_code):
-    with mock_client_call(mocker, 'add_composite_traintuple', response={}) as m:
+    with mock_client_call(mocker, 'add_composite_traintuple', response={}):
```
maeldebon: Looks like nothing is returned from `m`, that's why I removed it

tests/sdk/test_list.py
```diff
@@ -71,5 +71,5 @@ def test_list_asset_with_filters_failure(client, mocker):
     with pytest.raises(ValueError) as exc_info:
         client.list_algo(filters)
 
-    assert m.is_called()
+    m.assert_not_called()
```
maeldebon: Here, `m` should not be called, contrary to what was specified before 🙂

jmorel: 👍

tests/test_cli.py
jmorel: Actually there is something in m, however the issue is with the use of `with`. `mock_client_call` cannot be used as a context manager (see https://github.com/SubstraFoundation/substra/pull/41). You should use `m = mock_client_call(...)` and keep the assertion at the end.

Be careful through, half of the parametrization cases will trigger an error and therefore the mocker won't be called. You may need to either add an if before the assertion or split the test in 2, one for calls that success and the other for calls that fail.

tests/sdk/test_list.py
samlesu: 👍

samlesu:
Good catch @maeldebon 👍

tests/test_cli.py
```diff
@@ -197,28 +197,39 @@ def test_command_add_objective(workdir, mocker):
...
+])
+def test_command_add_composite_traintuple_missing_model_key(mocker, workdir, params, message):
+    m = mock_client_call(mocker, 'add_composite_traintuple', response={})
+    json_file = workdir / "valid_json_file.json"
+    json_file.write_text(json.dumps({}))
+    res = client_execute(workdir, ['add', 'composite_traintuple', '--algo-key', 'foo',
+                                   '--dataset-key', 'foo', '--data-samples-path', str(json_file)]
+                         + params, exit_code=2)
+    assert re.search(message, res)
+    m.assert_not_called()
```
jmorel: Since the mocker is never called, you could also just remove it entirely. 

The test on L224 already checks that we have the correct error, so even if it reached the non mocked add_composite_traintuple function, the test would fail because the connection failed error wouldn't match the `the <option> option is required` string passed as parameter.

maeldebon: That's right, I'll remove it!

PR #86, opened 2020-01-22
Update compute plan

jmorel:
Add support for an upcoming update compute plan endpoint, assuming:
- the endpoint is reachable by POST at `/compute_plan/<compute_plan_id>/update_ledger` (similar to update dataset)
- the add and update compute plan calls will return in their json a `keysToIDsMapping ` key that will contain the mapping

The mapping is then saved in a temp file and displayed if the call was made through the CLI.


jmorel:
As https://github.com/SubstraFoundation/substra-backend/pull/113, this feature isn't required anymore, I'm closing the PR.

PR #87, opened 2020-01-23
Change retry strategy on post request on ledger timeout

samlesu:
This is required as we don't know on a ledger timeout error if the request will be processed by the chaincode.

### DIFF
--- a/substra/sdk/rest_client.py
+++ b/substra/sdk/rest_client.py
@@ -207,34 +207,13 @@ class Client():
 
         return items
 
-    def add(self, name, retry_timeout=False, exist_ok=False, **request_kwargs):
-        """Add asset.
+    def _add(self, name, exist_ok=False, **request_kwargs):
+        """ Add asset wrapper.
 
-        In case of timeout, block till resource is created.
-
-        If `exist_ok` is true, `AlreadyExists` exceptions will be ignored and the
-        existing asset will be returned.
+        Handles conflict error when created asset already exists.
         """
         try:
-            return self.request(
-                'post',
-                name,
-                **request_kwargs,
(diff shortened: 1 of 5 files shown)


substra/sdk/rest_client.py
```diff
@@ -231,23 +248,13 @@ def add(self, name, retry_timeout=False, exist_ok=False, **request_kwargs):
...
-            key = e.pkhash
-            is_many = isinstance(key, list)
-            if is_many:
-                logger.warning("Many assets not compatible with 'exist_ok' option")
-                raise
-
-            logger.warning(f"{name} already exists: key='{key}'")
-            return self.get(name, key)
+            # XXX as there is no guarantee that the request has been sent to the ledger
+            #     (and will be processed), retry on on the add request and ignore
```
inalgnu:
```suggestion
            #     (and will be processed), retry on the add request and ignore
```

tests/sdk/utils.py
```diff
@@ -32,12 +32,13 @@ def mock_requests_response(response=None, status=200, headers=None):
...
-def mock_requests(mocker, method, response=None, status=200, headers=None):
-
-    def _req(*args, **kwargs):
-        return mock_requests_response(response, status, headers)
-
+def mock_requests_responses(mocker, method, responses):
     return mocker.patch(
         f'substra.sdk.rest_client.requests.{method}',
-        side_effect=_req,
+        side_effect=responses,
```
jmorel: It looks like there is only one use of this new function, but I see for example in `test_download.py` L72 that we used to do the same thing manually. I think we should either replace all mocks of requests using this function or none of them.

PR #88, opened 2020-01-24
Display compute plan progress in CLI

jmorel:
Since the compute plan now contains doneCount and tupleCount properties, we can display the overall progress to the user.

### DIFF
--- a/substra/cli/printers.py
+++ b/substra/cli/printers.py
@@ -116,13 +116,15 @@ class CurrentNodeField(Field):
         return ''
 
 
-class ComputePlanProgressField(Field):
-    def __init__(self, name):
+class ProgressField(Field):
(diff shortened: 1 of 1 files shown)


substra/cli/printers.py
```diff
@@ -116,6 +116,16 @@ def get_value(self, item, expand=False):
         return ''
 
 
+class ComputePlanProgressField(Field):
```
samlesu: What about creating a `ProgressField` with the following constructor for instance?

```python
def __init__(self, name, progress_ref, total_ref):
```

It will be more generic and could be used in other asset?

jmorel: Sure, why not!

jmorel: Done, thanks!

PR #91, opened 2020-01-28
Support multiple configs in examples

jmorel:
- adds a `SUBSTRA_EXAMPLES_CONFIG_FILEPATH ` variable env to select which backend config to use
- adds configs for both docker-compose and skaffold setups
- make skaffold the default config


samlesu:
On this topic I'm wondering if we could uses the existing `~/.subtra` config file instead of creating a new one. It seems wrong to create a new config file, we already have one.

Maybe we could ask the user to run the commands `substra config ... --profile node1` and  `substra config ... --profile node2` before running the example and then we could use in the examples directly:
```python
client = substra.Client(profile="node1")
client.login()
```

What do you think?

jmorel:
As much as I'd love to use such a solution, the fact that the password isn't stored in the profile will prevent it. The configs used in examples do store the password, this is why we can use them.

samlesu:
> As much as I'd love to use such a solution, the fact that the password isn't stored in the profile will prevent it. The configs used in examples do store the password, this is why we can use them.

The goal of the previous Users PR was to avoid storing the password on disk, and we are showing the opposite in our examples. It looks wrong to me.

If we tell the user to launch the command `substra config url ...` the password will be stored using keyring and users won't have to set it again / store it on disk. So I don't think it will be an issue.

jmorel:
True. I'll create a separate issue for this very nice improvement.

jmorel:
Closing this as the improvements described above have been implemented.

PR #95, opened 2020-01-29
Compatibility table

AurelienGasser:
This is a first attempt at having a central place where we can document which versions are compatible with each other.

Sometimes, as a dev, it's useful to have a working baseline that "we know works".

It can also be useful in deployment scenarios to document tested releases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,9 +91,11 @@ Learning about the Substra platform:
 
 These sets of versions have been tested for compatilibility: all the tests from [`substra-tests`](https://github.com/SubstraFoundation/substra-tests/) pass.
 
-| Major release?  | substra  | substra-chaincode  | substra-backend  | substra-tests  |
-|---|---|---|---|---|
-|   | [`b537d7`](https://github.com/SubstraFoundation/substra/commit/b537d7d3730dd74764213af9c93ee735005c5ac7)  | [`0.0.8-alpha.4`](https://github.com/SubstraFoundation/substra-chaincode/releases/tag/0.0.8-alpha.4)  | [`0.0.12-alpha.6`](https://github.com/SubstraFoundation/substra-backend/releases/tag/0.0.12-alpha.6) | [`fc52ea`](https://github.com/SubstraFoundation/substra-tests/commit/fc52ea1f7b2846e2af97d73d654b0265ded7011a) |
+| substra  | substra-chaincode  | substra-backend  | substra-tests  | hlf-k8s | substra-frontend (*) |
+|---|---|---|---|---|---|---|
+| [`b537d7`](https://github.com/SubstraFoundation/substra/commit/b537d7d3730dd74764213af9c93ee735005c5ac7) | [`0.0.8-alpha.4`](https://github.com/SubstraFoundation/substra-chaincode/releases/tag/0.0.8-alpha.4) | [`0.0.12-alpha.6`](https://github.com/SubstraFoundation/substra-backend/releases/tag/0.0.12-alpha.6) | [`fc52ea`](https://github.com/SubstraFoundation/substra-tests/commit/fc52ea1f7b2846e2af97d73d654b0265ded7011a) | [`9148fd`](https://github.com/SubstraFoundation/hlf-k8s/commit/9148fd349f72f5ce71619421616969581f7a9634) |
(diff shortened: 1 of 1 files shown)


AurelienGasser:
cc @Kelvin-M @GuillaumeCisco @ClementGautier FYI

samlesu:
Thanks @AurelienGasser. That's a very good idea!

1. What about adding a `hlf-k8s` column?
2. We could drop the column `major release`, we could add it once we exit the initial development mode (0.Y.Z).

README.md
```diff
@@ -86,6 +87,14 @@ Learning about the Substra platform:
 - [Cross-validation](./examples/cross_val/README.md)
 - [Compute plan](./examples/compute_plan/README.md)
 
+## Compatibility table
+
+These sets of versions have been tested for compatilibility: all the tests from [`substra-tests`](https://github.com/SubstraFoundation/substra-tests/) pass.
+
+| Major release?  | substra  | substra-chaincode  | substra-backend  | substra-tests  |
```
ClementGautier: I would remove the "major release" column as Samuel said and add the substra-frontend and hlf-k8s to the table.

AurelienGasser: Done.

substra-frontend is a bit trickier, because it doesn't have end to end tests. As a result, one can easily assume that the front end "works" after loading a few pages, when actually some other page is broken because of a recent backend/chaincode change. Adding a "not fully tested" substra-frontend version to the compatibility table would defeat the purpose of the compatibility table.

To mitigate this, I added a (*) linking to (yet to be written) frontend testing instructions, and instructing to leave the column blank unless the frontend has been fully tested.

Happy to have your thoughts!

samlesu: Looks good to me.

As far as I understand, we will add a new line to this table each time we release one of our "services". Is it correct? 

So for each new release, we should at least ensure the frontend is not broken.

AurelienGasser: > As far as I understand, we will add a new line to this table each time we release one of our "services". Is it correct?

I'm not sure which guidelines or rules we should have for adding a new row. Major releases seem like a minimum. We could do for minor releases too, or even commits without releases if we haven't released anything in a while :man_shrugging: I'm not sure. As long as we test thoroughly before adding a row, we could add rows pretty often!

> So for each new release, we should at least ensure the frontend is not broken.

Agree! However, in the absence of instructions, I don't think we currently have a way to do that reliably (except by people who are very familiar with the frontend and its testing @jmorel :eyes:)

ClementGautier: > As a result, one can easily assume that the front end "works" after loading a few pages, when actually some other page is broken because of a recent backend/chaincode change

I guess you don't test every function call or even all the API endpoints neither so that's ok for me to include the frontend even if we are not 100% sure: it will, for sure, help create a QA checklist on this repo too.

ClementGautier: EDIT: if you provide a test protocol for the frontend you should provide one for the other components (even if the protocol ends up being "launched this command and it returned 0") so everyone knows what to expect.

AurelienGasser: Good points!

AurelienGasser: https://github.com/SubstraFoundation/substra/pull/95/commits/a75c6dc561f502577a71081167bf492fee1d7c23 Does that look better?

ClementGautier:
> The idea behind is that I don't think it will be very interesting to know the compatibility from old commits, so sticking to old releases could be enough.

In my opinion, if you test a commit and were thinking about adding it to this table, better put a tag instead of putting the commit sha (so the docker is built if needed / the release process is triggered)

samlesu:
> In my opinion, if you test a commit and were thinking about adding it to this table, better put a tag instead of putting the commit sha (so the docker is built if needed / the release process is triggered)

Seems like a good rule 👍

AurelienGasser:
I created new tags for the untagged commits:

- https://github.com/SubstraFoundation/substra/releases/tag/0.4.0-alpha.2
- https://github.com/SubstraFoundation/substra-tests/releases/tag/0.1.1
- https://github.com/SubstraFoundation/hlf-k8s/releases/tag/0.0.10

I didn't create any release, just the tags. It's the first time I do this, please let me know if I did anything wrong :)

If all looks good, let's merge it!

PR #101, opened 2020-01-30
hotfix/#99 add objective_key to add_traintuple method

chrisalexandrepena:


jmorel:
Traintuples don't reference objectives anymore, only testtuple do. This change was introduced when we removed the automatic evaluation after training. You may be running an older version of the backend.

chrisalexandrepena:
indeed

PR #106, opened 2020-02-04
Fix rest client login headers

maeldebon:
🔗 **Related issue:** #100 
🔗 **Related PRs**:
- [Update examples with headers fix in rest_client](https://github.com/SubstraFoundation/substra/pull/108)
- [Update tests with headers fix in rest_client](https://github.com/SubstraFoundation/substra/pull/109)

This PR is possible thanks to the following fix in the backend: [#147](SubstraFoundation/substra-backend#147).

The objective of this PR is to be able to remove the configuration JSON files and to store the configuration directly in the headers.

PR #108, opened 2020-02-04
Update examples with headers fix in rest_client

maeldebon:
🔗 **Related PRs**:
- [Fix rest client login headers](https://github.com/SubstraFoundation/substra/pull/106)
- [Update tests with headers fix in rest_client](https://github.com/SubstraFoundation/substra/pull/109)

We are currently modifying the way the login headers works in #106. We are removing the config.json (where the password is set) and storing its content through the headers instead.
This PR has been opened in order to apply these changes in the examples we can find on the `substra` repository.
We are just replacing this:
```py
client = substra.Client()
client.add_profile(config['profile_name'], config['username'], config['password'],  config['url'])
client.login()
``` 

by this:
```py
client = substra.Client(profile_name="node-1")
client.login()
```

### DIFF
--- a/examples/compute_plan/scripts/add_compute_plan.py
+++ b/examples/compute_plan/scripts/add_compute_plan.py
@@ -26,7 +26,7 @@ compute_plan_keys_path = os.path.join(current_directory, '../compute_plan_keys.j
 with open(os.path.join(current_directory, '../../config.json'), 'r') as f:
     config = json.load(f)
 
-client = substra.Client()
+client = substra.Client(profile_name="node-1")
 client.login()
 
 print(f'Loading existing asset keys from {os.path.abspath(assets_keys_path)}...')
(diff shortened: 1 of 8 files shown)


examples/titanic/README.md
```diff
@@ -13,6 +13,11 @@ In order to run this example, you'll need to:
 * [install the `substra` cli](../../README.md#install)
 * [install the `substratools` library](https://github.com/substrafoundation/substra-tools)
 * [pull the `substra-tools` docker images](https://github.com/substrafoundation/substra-tools#pull-from-private-docker-registry)
+* have one or two profiles:
```
samlesu: I think we could improve the README by:
- giving more details regarding this step
- setting the URL from the skaffold setup (instead of docker-compose)

So something like the following lines could be better:

create the substra profile to define the substra network to target, for instance:
```sh
substra config --profile node-1 --username node-1 --password 'p@$swr0d44' http://substra-backend.node-1.com
```

examples/compute_plan/scripts/add_compute_plan.py
```diff
@@ -26,8 +26,7 @@
 with open(os.path.join(current_directory, '../../config.json'), 'r') as f:
     config = json.load(f)
```
samlesu: This code can be removed (and the config.json file could also be removed).

maeldebon: That's right 👍

jmorel:
👏 👏 👏  Very nice!

jmorel:
👏 👏 👏  Very nice!

PR #109, opened 2020-02-04
Update tests with headers fix in rest_client

maeldebon:
🔗 **Related PRs**:
- [Fix rest client login headers](https://github.com/SubstraFoundation/substra/pull/106)
- [Update examples with headers fix in rest_client](https://github.com/SubstraFoundation/substra/pull/108)

This PR aims to update the tests with the fixed headers in the rest client.

### DIFF
--- a/tests/sdk/test_config.py
+++ b/tests/sdk/test_config.py
@@ -48,33 +48,11 @@ def test_add_load_profile_from_file(tmpdir):
     }
 
     path.write_text(json.dumps(conf), "UTF-8")
-    keyring.set_password('node-1', conf['node-1']['auth']['username'], 'foo')
     manager = configuration.Manager(str(path))
     profile = manager.load_profile('node-1')
 
-    assert conf['node-1']['auth']['username'] == profile['auth']['username']
+    assert conf['node-1'] == profile
     assert keyring.get_password(service_name='node-1', username='node-1') == 'foo'
-
-
-def test_add_load_bad_profile_from_file(tmpdir):
-    path = tmpdir / 'substra.cfg'
-    conf = {
-       "node-1": {
-           "auth": {
-               "username": "node-1"
-           },
-           "insecure": False,
-           "url": "http://substra-backend.node-1.com",
-           "version": "0.0"
-       },
-    }
-
-    path.write_text(json.dumps(conf), "UTF-8")
-    keyring.set_password('node-1', conf['node-1']['auth']['username'], 'foo')
-    manager = configuration.Manager(str(path))
-
-    with pytest.raises(configuration.ProfileNotFoundError):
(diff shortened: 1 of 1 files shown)


tests/sdk/test_config.py
```diff
@@ -32,6 +34,49 @@ def test_add_load_profile(tmpdir):
...
+           "version": "0.0"
+       },
+    }
+
+    path.write_text(json.dumps(conf), "UTF-8")
+    keyring.set_password('node-1', conf['node-1']['auth']['username'], 'foo')
+    manager = configuration.Manager(str(path))
+    profile = manager.load_profile('node-1')
+
+    assert conf['node-1']['auth']['username'] == profile['auth']['username']
```
jmorel: Instead of checking just one property, I think you should be able to test all of them with something like 
```python
assert conf['node-1'] == profile
```

tests/sdk/test_config.py
```diff
@@ -32,6 +34,49 @@ def test_add_load_profile(tmpdir):
...
+    assert keyring.get_password(service_name='node-1', username='node-1') == 'foo'
```
jmorel: The keyring.set_password, keyring.get_password only test that keyring indeed works, but doesn't check anything from our code. When we're using keyring, we're assuming that it's been properly tested by its developers and don't need to re-test it.

tests/sdk/test_config.py
```diff
@@ -32,6 +34,49 @@ def test_add_load_profile(tmpdir):
...
+           "version": "0.0"
+       },
+    }
+
+    path.write_text(json.dumps(conf), "UTF-8")
+    keyring.set_password('node-1', conf['node-1']['auth']['username'], 'foo')
+    manager = configuration.Manager(str(path))
+
+    with pytest.raises(configuration.ProfileNotFoundError):
+        manager.load_profile('foo')
```
jmorel: Since the only test is about the error raised when trying to access a non-existing profile, I don't think that loading profiles from a file make any difference from loading the profiles with a python object. It's therefore redundant with the test on L89-L90.

You could however try to write a bad configuration to the file and see what error is being raised by `configuration.Manager(str(path))`

tests/sdk/test_config.py
maeldebon: You're right, it could be easier!

tests/sdk/test_config.py
maeldebon: Alright, so I should remove this link?

jmorel: I think so, yes.

tests/sdk/test_config.py
maeldebon: Yes, I'll do it 🙂

PR #120, opened 2020-02-18
Update compatibility matrix

samlesu:
The current (and only) entry is not working. This is due to the fact that the latest hlf-k8s release is deprecated.


AurelienGasser:
Do we want to delete the entry that doesn't work anymore?

samlesu:
> Do we want to delete the entry that doesn't work anymore?

Good idea. Done.

PR #110, opened 2020-02-04
[documentation] Update skaffold & docker-compose local install 

natct10:
This a proposition of documentation update that is more focused on Ubuntu (18.04 & 19.10), trying to take into account some VM issues. 

Please, feel free to raise any question, to point me towards any relevant resource, to give feedback or to suggest other reviewers!

It is related to the issue [#74](https://github.com/SubstraFoundation/substra/issues/74)
Fix #74 
Fix #96

PR #121, opened 2020-02-21
Add clear error message when sdk can't fetch from keyring

### LINKED ISSUE #75: Keyring might not return the store password
Keyring may not be working on some OS (seen on Debian). 

One encountered error is that it is failing silently to fetch the password. It leads then to an authentication error, this is misleading for the user.

To solve this, we could:
- display a proper error message when keyring fails to fetch the password (Fetching password error: please check your keyring installation)
- update the documentation for Debian (requires https://packages.debian.org/fr/stretch/debian-keyring and https://pypi.org/project/keyrings.alt/) (optional)

AlexandrePicosson:
resolve #75 

The documentation part of this issue is covered by https://github.com/SubstraFoundation/substra/pull/110.

### DIFF
--- a/substra/sdk/rest_client.py
+++ b/substra/sdk/rest_client.py
@@ -89,9 +89,16 @@ class Client():
             raise exceptions.BadConfiguration('Your configuration is outdated, please update it.')
 
         username = config['auth']['username']
+        password = keyring.get_password(profile_name, username)
+
+        if password is None:
+            raise exceptions.KeyringException(
+                Exception("Fetching password error: Check your keyring installation")
+            )
+
         self._auth = {
             'username': username,
-            'password': keyring.get_password(profile_name, username)
+            'password': password
         }
 
     def __request(self, request_name, url, **request_kwargs):
--- a/substra/sdk/rest_client.py
+++ b/substra/sdk/rest_client.py
@@ -93,7 +93,7 @@ class Client():
 
         if password is None:
             raise exceptions.KeyringException(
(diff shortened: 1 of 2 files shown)


substra/sdk/rest_client.py
```diff
@@ -89,9 +89,16 @@ def set_config(self, config, profile_name='default'):
             raise exceptions.BadConfiguration('Your configuration is outdated, please update it.')
 
         username = config['auth']['username']
+        password = keyring.get_password(profile_name, username)
+
+        if password is None:
+            raise exceptions.KeyringException(
+                Exception("Fetching password error: Check your keyring installation")
```
samlesu:
```suggestion
               "Fetching password error: Check your keyring installation"
```

samlesu: Is there a reason for creating an exception inside an exception?

AlexandrePicosson: No, i'm just not familiar with the exceptions in python and thought it was the right way to do it

substra/sdk/rest_client.py
```diff
@@ -89,9 +89,16 @@ def set_config(self, config, profile_name='default'):
             raise exceptions.BadConfiguration('Your configuration is outdated, please update it.')
 
         username = config['auth']['username']
+        password = keyring.get_password(profile_name, username)
+
+        if password is None:
+            raise exceptions.KeyringException(
+                'Fetching password error: Check your keyring installation'
+            )
```
natct10: :+1:

PR #123, opened 2020-02-24
add dockerfile

AurelienGasser:
Companion PR: https://github.com/SubstraFoundation/substra-tests/pull/81

^ see companion PR for details

PR #123 :sunglasses:

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,12 @@
+FROM python:3.7
+
+RUN pip3 install keyrings.alt
+
+RUN mkdir -p /usr/src/app/substra
+COPY ./requirements.txt /usr/src/app/substra/
+COPY ./substra/ /usr/src/app/substra/substra/
+COPY ./setup.* /usr/src/app/substra/
+COPY ./README.md /usr/src/app/substra/
+RUN pip3 install -e /usr/src/app/substra
+
+WORKDIR /usr/src/app
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
-.PHONY: pyclean doc test
+DOCKER_TAG := latest
+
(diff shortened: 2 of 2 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN pip3 install keyrings.alt
+
+RUN mkdir -p /usr/src/app/substra
+COPY ./requirements.txt /usr/src/app/substra/
+COPY ./substra/ /usr/src/app/substra/substra/
+COPY ./setup.* /usr/src/app/substra/
+COPY ./README.md /usr/src/app/substra/
+RUN pip3 install -e /usr/src/app/substra
+
+WORKDIR /usr/src/app
```
ClementGautier: I would have put the workdir on front so you can use `.` as context. You are copying almost all files here so you can probably simplify a little bit the copy part and the the pip install should be placed before the source copy so that you can benefit from docker cache layers. Also, why is `keyrings.alt` not part of the requirements if it's needed? In my opinion the Dockerfile should look like this:

```
FROM python:3.7

WORKDIR /usr/src/app

COPY requirements.txt .
RUN pip install -e --no-cache-dir -r requirements.txt

COPY . .

CMD [ "substra" ]
```

AlexandrePicosson: I think keyrings.alt is needed only if the default OS keyring is not supported so we don't want to add it as a requirement for substra.

AurelienGasser: @AlexandrePicosson There's no default OS keyring on this base image, and all my attempts to make one work have failed!

AurelienGasser:
Closing this as it's [not needed anymore](https://github.com/SubstraFoundation/substra-tests/pull/81#issuecomment-591059782).

PR #125, opened 2020-03-02
Add new entry to compatibility matrice

samlesu:


inalgnu:
for the chaincode I've just tagged a new version 0.0.8-alpha.10 should be compatible with rest, adding a small bug fix

Issue #114, opened 2020-02-06
Create an aggregatetuple without in model key


maeldebon:
It is possible to create an aggregatetuple without passing any --in-model-key in the command.

jmorel:
@maeldebon following the short discussion we've had in the PR, I'm moving this issue to the backend repository.

PR #127, opened 2020-03-04
Cannot add aggregatetuple without in-model key(s)

maeldebon:
🔗 **This PR is related to this issue: #114**

Currently, this is possible to add an aggregatetuple without passing the `--in-model-key` argument.
The objective of this PR is to avoid being able to do so.

### DIFF
--- a/references/cli.md
+++ b/references/cli.md
@@ -411,7 +411,7 @@ Usage: substra add aggregatetuple [OPTIONS]
 
 Options:
   --algo-key TEXT                 [required]
-  --in-model-key TEXT             In model traintuple key.
+  --in-model-key TEXT             In model traintuple key.  [required]
   --worker TEXT                   Node ID for worker execution.  [required]
   --rank INTEGER
   --tag TEXT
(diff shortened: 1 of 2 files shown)


jmorel:
Could you show what happened before your change if you added an aggregatetuple without inmodels? Was the call rejected?

maeldebon:
> Could you show what happened before your change if you added an aggregatetuple without inmodels? Was the call rejected?

No, the aggregatetuple was successfully created, with the normal output:
```
KEY                 e4b6864bbc897b7db8dad988d7b86e6fc37051ebecf832e485feb462ed5fc44b
MODEL KEY           None
ALGO KEY            0099f052d872348c91872da792df9c86f2546b8b8204c5aa5881b708f0639319
ALGO NAME           a94d7c67917a412e807908cb7661c086_test_aggregatetuple - Algo 1
STATUS              todo
DATASET KEY         None
IN MODEL KEYS       None
RANK                0
COMPUTE PLAN ID
TAG
LOG
CREATOR             MyOrg1MSP
WORKER              None
PERMISSIONS         Processable by anyone

Display this aggregatetuple's testtuples:
	substra list testtuple -f "testtuple:traintupleKey:e4b6864bbc897b7db8dad988d7b86e6fc37051ebecf832e485feb462ed5fc44b"
```

jmorel:
I don't think we should add validation in the CLI that is not also enforced on the backend side. As long as the call is accepted, there is no reason to make the option mandatory.

maeldebon:
> I don't think we should add validation in the CLI that is not also enforced on the backend side. As long as the call is accepted, there is no reason to make the option mandatory.

Okay! 👍

PR #128, opened 2020-03-04
Drop Python 2 support

### LINKED ISSUE #119: Drop support for python < 3
Python 2.6 and 2.7 are currently not tested and I think we should drop its support. 

To do so, we should update the documentation and the setup.py.

natct10:
This PR intends at dropping support for Python 2 as it not maintained anymore; this could help newcomers to the project avoid some issues. 
The documentation part of this modification is to be found in the ongoing documentation refresh [PR](https://github.com/SubstraFoundation/substra/pull/110#). 
Do you think any extra steps are required?

Fixes #119


samlesu:
Thanks @natct10 !

I'm wondering if we should add a `python_requires` field to the setup.py file? (cf https://packaging.python.org/guides/dropping-older-python-versions/#specify-the-version-ranges-for-supported-python-distributions)

jmorel:
It looks like there are quite a few f-strings in the codebase and those were introduced in python3.6. Shouldn't we also specify that earlier version won't work?

natct10:
> It looks like there are quite a few f-strings in the codebase and those were introduced in python3.6. Shouldn't we also specify that earlier version won't work?

Good point @jmorel !

natct10:
I have updated the documentation accordingly; please let me know if anything else is required!

PR #133, opened 2020-03-06
Add algo type in add aggregatetuple

maeldebon:
Currently, when using the following command, the type of the `--algo-key` option isn't specified: 
```sh
substra add aggregatetuple --help
```

Maybe we could add this information in order to make this doc easier to understand?


maeldebon:
What do you think about this @samlesu?

samlesu:
> What do you think about this @samlesu?

Good idea :)

PR #135, opened 2020-03-11
Add clean model and tag option in the documentation

Kelvin-M:
Companion PR https://github.com/SubstraFoundation/substra-backend/pull/172


jmorel:
Could you also add the `clean_model` field to the ComputePlanPrinter (in both list and detail field lists) ?

PR #138, opened 2020-03-13
Remove use of keyring

### LINKED ISSUE #97: Error when trying to run substra config
I am trying to run the `substra config http://substra-backend.node-1.com:80 -v 0.0 -u node-1 -p 'p@$swr0d44'` command on my ubuntu server 18.04.3 LTS environment. Here is the error I get :

```
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/bin/substra", line 8, in <module>
    sys.exit(cli())
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/click/core.py", line 956, in invoke
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/click/core.py", line 555, in invoke
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/substra/cli/interface.py", line 256, in add_profile_to_config
    insecure=insecure,
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/substra/sdk/config.py", line 125, in add_profile
    keyring.set_password(name, username, password)
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/keyring/core.py", line 62, in set_password
    _keyring_backend.set_password(service_name, username, password)
  File "/home/ubuntu/dev/substra/examples/titanic/.venv/lib/python3.6/site-packages/keyring/backends/fail.py", line 24, in get_password
    raise RuntimeError(msg)
RuntimeError: No recommended backend was available. Install a recommended 3rd party backend package; or, install the keyrings.alt package if you want to use the non-recommended backends. See https://pypi.org/project/keyring for details.
```

The error gets fixed if I install the keyrings.alt dependency.

### LINKED ISSUE #165: keyrings.alt requirement necessary for Linux machines
When registering new samples with substra on a Linux machine the error `samples:keyrings.error.NoKeyRingsError` sometimes forces the user to install `keyrings.alt`. 
But it seems like it is might not be a recommended solution, any idea on that ?

jmorel:
This PR removes the storage of user passwords in the keyring. 

The client can now be initialized with an existing token or will use the one from `~/.substra-user`.

## CLI changes

- removed `--username` and `--password` options from the `config` command
- added `--username` and `--password` options to the `login` command. If not provided at call, the command will prompt the user for values and the password won't be displayed at all.

## SDK changes

- the `Client` constructor takes an optional `token` arg instead of the previous `username` and `password`
- `add_profile` doesn't have `username` and `password` anymore
- new `set_token` method

## Companion PRs

Since this introduces a change in our public interface, there are a few companion PRs:
* substra-backend: https://github.com/SubstraFoundation/substra-backend/pull/202
* substra-tests: https://github.com/SubstraFoundation/substra-tests/pull/94
* ~hlf-k8s: https://github.com/SubstraFoundation/hlf-k8s/pull/42~
* substra-documentation: https://github.com/SubstraFoundation/substra-documentation/pull/53

Fixes #97 (makes it irrelevant)
Fixes #165

### DIFF
--- a/examples/titanic/README.md
+++ b/examples/titanic/README.md
@@ -15,8 +15,8 @@ In order to run this example, you'll need to:
 * [pull the `substra-tools` docker images](https://github.com/substrafoundation/substra-tools#pull-from-private-docker-registry)
 * create a substra profile to define the substra network to target, for instance:
     ```sh
-    substra config --profile node-1 --username node-1 --password 'p@$swr0d44' http://substra-backend.node-1.com
-    substra login --profile node-1
+    substra config --profile node-1 http://substra-backend.node-1.com
+    substra login --profile node-1 --username node-1 --password 'p@$swr0d44'
     ```
 * checkout this repository
 
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -265,14 +265,10 @@ def cli(ctx):
 @click.option('--insecure', '-k', is_flag=True,
               help='Do not verify SSL certificates')
 @click.option('--version', '-v', default=configuration.DEFAULT_VERSION)
-@click.option('--username', '-u', required=True)
-@click.option('--password', '-p', required=True)
-def add_profile_to_config(url, config, profile, insecure, version, username, password):
+def add_profile_to_config(url, config, profile, insecure, version):
     """Add profile to config file."""
     configuration.Manager(config).add_profile(
         profile,
-        username,
-        password,
         url,
         version=version,
         insecure=insecure,
@@ -283,12 +279,14 @@ def add_profile_to_config(url, config, profile, insecure, version, username, pas
 @click_global_conf
 @click.pass_context
 @error_printer
-def login(ctx):
+@click.option('--username', '-u', prompt=True)
+@click.option('--password', '-p', prompt=True, hide_input=True)
+def login(ctx, username, password):
     """Login to the Substra platform."""
     usr.Manager(ctx.obj.user).clear_user()
     client = get_client(ctx.obj)
 
-    token = client.login()
+    token = client.login(username, password)
     # create temporary user data
     usr.Manager(ctx.obj.user).add_user(token)
 
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -279,8 +279,8 @@ def add_profile_to_config(url, config, profile, insecure, version):
 @click_global_conf
 @click.pass_context
 @error_printer
-@click.option('--username', '-u', prompt=True)
-@click.option('--password', '-p', prompt=True, hide_input=True)
+@click.option('--username', '-u', envvar='SUBSTRA_USERNAME', prompt=True)
+@click.option('--password', '-p', envvar='SUBSTRA_PASSWORD', prompt=True, hide_input=True)
 def login(ctx, username, password):
     """Login to the Substra platform."""
     usr.Manager(ctx.obj.user).clear_user()
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -289,6 +289,7 @@ def login(ctx, username, password):
     token = client.login(username, password)
     # create temporary user data
     usr.Manager(ctx.obj.user).add_user(token)
+    display(f"Token: {token}")
 
 
 @cli.group()
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -19,8 +19,6 @@ import os
 import time
 import json
 
-import keyring
-
 from substra.sdk import utils, assets, rest_client, exceptions
 from substra.sdk import config as cfg
 from substra.sdk import user as usr
@@ -57,7 +55,7 @@ def get_asset_key(data):
 class Client(object):
 
     def __init__(self, config_path=None, profile_name=None, user_path=None,
-                 retry_timeout=DEFAULT_RETRY_TIMEOUT):
+                 token=None, retry_timeout=DEFAULT_RETRY_TIMEOUT):
         self._cfg_manager = cfg.Manager(config_path or cfg.DEFAULT_PATH)
         self._usr_manager = usr.Manager(user_path or usr.DEFAULT_PATH)
         self._current_profile = None
@@ -73,15 +71,18 @@ class Client(object):
         # set current logged user if exists
         self.set_user()
 
+        if token:
+            self.set_token(token)
+
     @logit
-    def login(self):
+    def login(self, username, password):
         """Login.
 
         Allow to login to a remote server.
 
-        After setting your configuration with `substra config` using `-u` and `-p`
-        Launch `substra login`
-        You will get a token which will be stored by default in `~/.substra-user`
+        After setting your configuration with `substra config`, launch `substra login`.
+        You will be prompted for your username and password and get a token which will be
+        stored by default in `~/.substra-user`
         You can change that thanks to the --user option (works like the --profile option)
 
         """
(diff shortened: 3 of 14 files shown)


substra/cli/interface.py
```diff
@@ -278,12 +274,14 @@ def add_profile_to_config(url, config, profile, insecure, version, username, pas
 @click_global_conf
 @click.pass_context
 @error_printer
-def login(ctx):
+@click.option('--username', '-u', prompt=True)
+@click.option('--password', '-p', prompt=True, hide_input=True)
+def login(ctx, username, password):
```
samlesu: This command could print the token ?

substra/sdk/client.py
```diff
@@ -97,24 +95,34 @@ def __init__(self, config_path=None, profile_name=None, user_path=None,
...
         if not self._current_profile:
             raise exceptions.SDKException("No profile defined")
 
-        res = self.client.login()
+        res = self.client.login(username, password)
         token = res.json()['token']
+        self.set_token(token)
+        return token
+
+    def set_token(self, token):
```
samlesu: This method could be private as it's not needed to the end user. 

Having smaller public interfaces is easier to maintain (we can do more changes without breaking the current interface).

substra/sdk/client.py
```diff
@@ -97,24 +95,34 @@ def __init__(self, config_path=None, profile_name=None, user_path=None,
...
 
         Allow to login to a remote server.
 
-        After setting your configuration with `substra config` using `-u` and `-p`
-        Launch `substra login`
-        You will get a token which will be stored by default in `~/.substra-user`
+        After setting your configuration with `substra config`, launch `substra login`.
+        You will be prompted for your username and password and get a token which will be
+        stored by default in `~/.substra-user`
         You can change that thanks to the --user option (works like the --profile option)
```
samlesu: Not related directly to this PR, but this doctring seems related to the CLI (and not the SDK).

substra/sdk/client.py
```diff
@@ -97,24 +95,34 @@ def __init__(self, config_path=None, profile_name=None, user_path=None,
         # set current logged user if exists
         self.set_user()
 
+        if token:
+            self.set_token(token)
+
     @logit
-    def login(self):
+    def login(self, username, password):
```
samlesu: Could be renamed to authenticate? It could be worth to investigate what's the best / more common verb to describe this action.

substra/cli/interface.py
jmorel: Sure, good idea!

substra/sdk/client.py
jmorel: Sure!

substra/sdk/client.py
jmorel: This would break the matching between names in the CLI and names in the SDK though.

substra/sdk/client.py
jmorel: Agreed, I'll change it.

substra/sdk/client.py
samlesu: Yes, I was thinking to maybe rename both but I don't think it's worth it.

examples/titanic/README.md
```diff
@@ -15,19 +15,19 @@ In order to run this example, you'll need to:
...
-    substra login --profile node-1
+    substra config --profile node-1 http://substra-backend.node-1.com
+    substra login --profile node-1 --username node-1 --password 'p@$swr0d44'
     ```
 
 ## Data preparation
 
-The first step will be to generate train and test data samples from the 
-[Kaggle challenge source](https://www.kaggle.com/c/titanic/data). The raw `train.csv` file is available in the `data` 
+The first step will be to generate train and test data samples from the
```
samlesu: To avoid adding a lot of unwanted changes (done by your editor I guess), you could use the `git add -p` to select only the desired changes.

substra/sdk/client.py
```diff
@@ -81,7 +79,7 @@ def _update_permissions_field(data, permissions_field='permissions'):
 class Client(object):
 
     def __init__(self, config_path=None, profile_name=None, user_path=None,
-                 retry_timeout=DEFAULT_RETRY_TIMEOUT):
+                 token=None, retry_timeout=DEFAULT_RETRY_TIMEOUT):
```
samlesu: The input token seems unused here.

substra/sdk/client.py
```diff
@@ -97,24 +95,29 @@ def __init__(self, config_path=None, profile_name=None, user_path=None,
...
-        After setting your configuration with `substra config` using `-u` and `-p`
-        Launch `substra login`
-        You will get a token which will be stored by default in `~/.substra-user`
-        You can change that thanks to the --user option (works like the --profile option)
+    @logit
+    def login(self, username, password):
+        """Login to a remote server.
 
+        Will log into the current profile's server and save the resulting token in the configured
+        user_path.
```
samlesu: At first sight it looks like the resulting token is not stored in the user_path. Correct me if I'm wrong, I think this is valid only for the CLI login command.

examples/titanic/README.md
samlesu: There is also an option on github to not display these changes. Forgot my previous message.

substra/sdk/client.py
jmorel: Yes you're right, I'll update the docstring.

substra/sdk/client.py
jmorel: Ah sorry, I'll remove it right away.

jmorel: Sorry*2, it is used, on lines 98-99

samlesu: It could be useful to accept a token as an input variable; maybe we could update the constructor method to handle it?

samlesu: 👍 sorry ! I have been confused with the diff.

AurelienGasser:
I think there's a potential companion PR here: https://github.com/SubstraFoundation/substra-tests/blob/master/docker/Dockerfile#L5

natct10:
> I love these changes!
> 
> A possible addition for your consideration: the ability to set the password from an environment variable, [à la pgsql](https://stackoverflow.com/a/6405296/1370722)
> 
> ```shell
> export SUBSTRA_PASSWORD=foo
> substra login --profile bar --user baz
> ```
> 
> This can prevent the password to be leaked in some cases, for instance
> 
> ```shell
> #!/bin/bash
> 
> set -x # Echo commands and expand variables
> 
> # Password leaked in the logs
> substra login --profile bar --user baz --password $SUBSTRA_PASSWORD
> 
> # Password not leaked in the logs. The env variable was set prior to running this script.
> substra login --profile bar --user baz
> ```

Neat!

jmorel:
@AurelienGasser thanks for the suggestion, it was actually super easy to do using click. This library really does it all!

And very good point about the companion PR in substra-tests. I'm realizing that actually there should be companion PRs in multiple repos, I'll update the main description of this PR with links.

samlesu:
Has been rebased on top of the master branch.

PR #140, opened 2020-03-17
Fix titanic readme following removal of auto eval

jmorel:
The commands from the `Using asset command line interfaces` section assumed there was a prediction after each train, which there isn't anymore. I added the predict command and fixed the local of all output files.

### DIFF
--- a/examples/titanic/README.md
+++ b/examples/titanic/README.md
@@ -21,13 +21,13 @@ In order to run this example, you'll need to:
 
 ## Data preparation
 
-The first step will be to generate train and test data samples from the 
-[Kaggle challenge source](https://www.kaggle.com/c/titanic/data). The raw `train.csv` file is available in the `data` 
+The first step will be to generate train and test data samples from the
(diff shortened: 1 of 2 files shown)


examples/titanic/README.md
```diff
@@ -70,13 +70,25 @@ python assets/algo_random_forest/algo.py train \
...
+  model
+
 python assets/objective/metrics.py \
   --debug \
   --opener-path assets/dataset/opener.py \
   --data-samples-path assets/train_data_samples \
-  --input-predictions-path assets/pred-train.csv
+  --input-predictions-path assets/pred-train.csv \
+  --output-perf-path assets/perf-train.json \
+  --log-path assets/logs/train_metrics.log
```
samlesu: As far as I understand the prediction and the score should not be part of the training task?

jmorel: The training score can be interesting, but it's not automatically computed anymore.

PR #143, opened 2020-03-19
Add user metadata

maeldebon:
🔗 **Related PRs:**
- [substra-tests](https://github.com/SubstraFoundation/substra-tests/pull/90)
- [substra-chaincode](https://github.com/SubstraFoundation/substra-chaincode/pull/111)
- [substra-backend](https://github.com/SubstraFoundation/substra-backend/pull/186)
- [substra-frontend (coming soon)](#)

The goal of these PRs is to add a `metadata` field in every asset in order to be able to store information related to this asset. It could be, for instance, a creation_ts, an epoch, a project_name, ...

### DIFF
--- a/references/cli.md
+++ b/references/cli.md
@@ -108,6 +108,7 @@ Usage: substra add dataset [OPTIONS] PATH
           "public": bool,
           "authorized_ids": list[str],
       },
+      "metadata": dict
   }
 
   Where:
@@ -152,6 +153,7 @@ Usage: substra add objective [OPTIONS] PATH
           "public": bool,
           "authorized_ids": list[str],
       },
+      "metadata": dict
   }
 
   Where:
@@ -205,6 +207,7 @@ Usage: substra add algo [OPTIONS] PATH
           "public": bool,
           "authorized_ids": list[str],
       },
+      "metadata": dict
   }
 
   Where:
@@ -271,7 +274,8 @@ Usage: substra add compute_plan [OPTIONS] TUPLES_PATH
           "test_data_sample_keys": list[str],
           "traintuple_id": str,
           "tag": str,
-      }]
+      }],
+      "metadata": list[str]
   }
 
 Options:
@@ -305,6 +309,7 @@ Usage: substra add aggregate_algo [OPTIONS] PATH
           "public": bool,
           "authorized_ids": list[str],
       },
+      "metadata": dict
   }
 
   Where:
@@ -345,6 +350,7 @@ Usage: substra add composite_algo [OPTIONS] PATH
           "public": bool,
           "authorized_ids": list[str],
       },
+      "metadata": dict
   }
 
   Where:
@@ -401,6 +407,7 @@ Options:
   --verbose                       Enable verbose mode.
   -o, --output [pretty|yaml|json]
                                   Set output format  [default: pretty]
+  --metadata-path FILE            Metadata file path
   --help                          Show this message and exit.
 ```
 
@@ -427,6 +434,7 @@ Options:
   --verbose                       Enable verbose mode.
   -o, --output [pretty|yaml|json]
                                   Set output format  [default: pretty]
+  --metadata-path FILE            Metadata file path
   --help                          Show this message and exit.
 ```
 
@@ -473,6 +481,7 @@ Options:
   --verbose                       Enable verbose mode.
   -o, --output [pretty|yaml|json]
                                   Set output format  [default: pretty]
+  --metadata-path FILE            Metadata file path
   --help                          Show this message and exit.
 ```
 
@@ -509,6 +518,7 @@ Options:
   --verbose                       Enable verbose mode.
   -o, --output [pretty|yaml|json]
                                   Set output format  [default: pretty]
+  --metadata-path FILE            Metadata file path
   --help                          Show this message and exit.
 ```
 
--- a/references/cli.md
+++ b/references/cli.md
@@ -274,8 +274,7 @@ Usage: substra add compute_plan [OPTIONS] TUPLES_PATH
           "test_data_sample_keys": list[str],
           "traintuple_id": str,
           "tag": str,
-      }],
-      "metadata": list[str]
+      }]
   }
 
 Options:
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -531,8 +531,7 @@ def add_compute_plan(ctx, tuples):
             "test_data_sample_keys": list[str],
             "traintuple_id": str,
             "tag": str,
-        }],
-        "metadata": list[str]
+        }]
     }
 
     """
--- a/substra/cli/interface.py
+++ b/substra/cli/interface.py
@@ -167,6 +167,16 @@ def click_global_conf_with_output_format(f):
     return f
 
 
+def click_option_metadata(f):
+    """Add metadata option to command."""
+    return click.option(
+        '--metadata-path',
+        'metadata',
+        callback=load_json_from_path,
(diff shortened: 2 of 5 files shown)


substra/cli/interface.py
```diff
@@ -112,6 +112,15 @@ def click_global_conf_config(f):
         help='Config path (default ~/.substra).')(f)
 
 
+def click_global_conf_metadata(f):
+    """Add metadata option to command."""
+    return click.option(
+        '--metadata',
+        callback=update_global_conf,
```
jmorel: Metadata is an option that is only available for `add` commands, and it is forwarded to the API. The global conf option on the other hand is here to tell the cli how to behave locally (output format, which conf to use etc.) I don't think they should be mixed together.

substra/cli/interface.py
```diff
@@ -112,6 +112,15 @@ def click_global_conf_config(f):
...
+        type=click.Path(exists=True, resolve_path=True),
+        help='Metadata file path')(f)
```
jmorel: Since metadata is a path, you may want to name and use it the same way we did for the other paths (`--data-samples-path` and `--out-trunk-model-permissions-path`): name it `--metadata-path`  and use the `load_json_from_path` callback.

substra/sdk/client.py
```diff
@@ -291,6 +291,7 @@ def add_dataset(self, data, exist_ok=False):
                 "public": bool,
                 "authorized_ids": list[str],
             },
+            "metadata": list[str]
```
jmorel: `list[str]` means that acceptable values are like `['a', 'b', 'c']`. Instead we want acceptable values to be `{'a': 'b', 'c': 'd'}`.

In python typing, this is a Dict and not a List : https://docs.python.org/3/library/typing.html#typing.Dict

jmorel:
@maeldebon could you edit the "coming soon" out of the link to the backend PR (and the same think in the backend PR as well)?

substra/sdk/client.py
maeldebon: Yes, you're right!

substra/cli/interface.py
```diff
@@ -341,9 +351,10 @@ def add_data_sample(ctx, path, dataset_key, local, multiple, test_only):
                 metavar="PATH")
 @click.option('--objective-key')
 @click_global_conf_with_output_format
+@click_option_metadata
 @click.pass_context
 @error_printer
-def add_dataset(ctx, data, objective_key):
+def add_dataset(ctx, data, objective_key, metadata):
```
samlesu: I think it would be better to not add an additional path but add a metadata field in the existing data file (same for all assets).

substra/cli/interface.py
```diff
@@ -186,6 +186,16 @@ def click_global_conf_with_output_format(f):
     return f
 
 
+def click_option_metadata(f):
+    """Add metadata option to command."""
+    return click.option(
+        '--metadata-path',
+        'metadata',
+        callback=load_json_from_path,
+        type=click.Path(exists=True, resolve_path=True),
```
jmorel: Could you add `dir_okay=False` to click.Path?

jmorel:
After looking at the substra-tests PR I realize you forgot to add metadata to compute plans.

substra/sdk/client.py
```diff
@@ -268,6 +289,7 @@ def add_data_samples(self, data, local=True):
         better control in case of failures.
         """
         data = _update_permissions_field(data)
+        data = _update_metadata_field(data)
```
maeldebon: We could maybe avoir having this by creating a common method for these cases?

substra/sdk/client.py
```diff
@@ -53,6 +53,26 @@ def get_asset_key(data):
     return data.get('pkhash') or data.get('key')
 
 
+def _flatten_metadata(data, field_name):
+    data = deepcopy(data)
+
+    for i in data[field_name]:
+        data[f'{field_name}_{i}'] = data[field_name].get(i, [])
```
jmorel: Since you're already iterating on the keys of `data[field_name]`, you are sure that data[field_name][i] is defined, you don't need to use `.get()`.

substra/sdk/client.py
```diff
@@ -53,6 +53,26 @@ def get_asset_key(data):
...
+
+    del data[field_name]
+    return data
+
+
+def _update_metadata_field(data, metadata_field='metadata'):
```
jmorel: Since metadata_field is never used, you can hardcode 'metadata' everywhere.

substra/cli/interface.py
```diff
@@ -521,7 +536,8 @@ def add_compute_plan(ctx, tuples):
             "test_data_sample_keys": list[str],
             "traintuple_id": str,
             "tag": str,
-        }]
+        }],
+        "metadata": list[str]
```
jmorel, line 540: There are also metadata for compute plan's tuples, they need to be included here as well.

substra/sdk/client.py
```diff
@@ -225,6 +247,7 @@ def add_data_sample(self, data, local=True, exist_ok=False):
 
         """
         data = _update_permissions_field(data)
+        data = _update_metadata_field(data)
```
jmorel, line 250: data samples don't have metadata.

substra/sdk/client.py
```diff
@@ -268,6 +291,7 @@ def add_data_samples(self, data, local=True):
         better control in case of failures.
         """
         data = _update_permissions_field(data)
+        data = _update_metadata_field(data)
```
jmorel, line 294: data samples don't have metadata.

substra/sdk/client.py
```diff
@@ -571,7 +612,8 @@ def add_compute_plan(self, data):
                 "test_data_sample_keys": list[str],
                 "traintuple_id": str,
                 "tag": str,
-            }]
+            }],
+            "metadata": list[str]
```
jmorel, line 616: you're missing the _update_metadata_field here. Also, we'll need to update the metadata field in each tuple of the compute plan.

substra/cli/interface.py
jmorel: Also, metadata is not a list[str]

references/cli.md
```diff
@@ -271,7 +274,8 @@ Usage: substra add compute_plan [OPTIONS] TUPLES_PATH
           "test_data_sample_keys": list[str],
           "traintuple_id": str,
           "tag": str,
-      }]
+      }],
+      "metadata": list[str]
```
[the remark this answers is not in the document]
maeldebon: That's probably a mistake, I'll change this!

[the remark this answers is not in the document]
maeldebon: Well, this is related to compute_plan, I should maybe remove it for now?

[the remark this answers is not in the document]
samlesu: Yes, if it's not available, we should not document it. This would be misleading for our users.

substra/sdk/client.py
```diff
@@ -583,6 +592,7 @@ def add_compute_plan(self, data):
             }],
             "clean_models": bool,
             "tag": str
+            "metadata": list[str]
```
jmorel, line 595: I believe this is a leftover that you should remove.

jmorel:
I'm writing down here what's been decided: compute plan metadata and metadata in printers will be done in separate PRs.

substra/sdk/client.py
maeldebon: Indeed, I forgot to remove it!

jmorel:
@maeldebon you need to re-generate the references.

jmorel:
@maeldebon flake8 is failing :(

PR #144, opened 2020-03-20
Document when AlreadyExists exception are raised

### LINKED ISSUE #92: Doc: add asset unicity rules in documentation
It can be quite frustrating to know what must be modified in order to push a new version of an asset.

For instance to fix an objective metrics Dockerfile, the user must also update the associated description.

jmorel:
This branch adds documentation to `add_` methods in the client specifying when the `AlreadyExists` exception are raised.

To make the generation of documentation a bit faster when testing, I to the makefile two new targets: `doc-cli` and `doc-sdk`.

Fixes #92

### DIFF
--- a/references/sdk.md
+++ b/references/sdk.md
@@ -173,7 +173,8 @@ Create new algo asset.
 }
 ```
 
-If an algo with the same file already exists, an `AlreadyExists` exception will be raised.
+If an algo with the same archive file already exists, an `AlreadyExists` exception will be
+raised.
 
 If `exist_ok` is true, `AlreadyExists` exceptions will be ignored and the
 existing asset will be returned.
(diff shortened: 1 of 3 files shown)


references/sdk.md
```diff
@@ -161,6 +173,8 @@ Create new algo asset.
 }
 ```
 
+If an algo with the same file already exists, an `AlreadyExists` exception will be raised.
```
samlesu:
```suggestion
If an algo with the same archive file already exists, an `AlreadyExists` exception will be raised.
```

I'm afraid people could get confused between the archive file vs the algo script file (same for other algos). What do you think?

jmorel: Good point, I'll make the update (and for objective as well).

jmorel: Updated. As for the objective I was mistaken as it's the description that is relevant and not the archive.

PR #146, opened 2020-03-20
Type hints

### LINKED ISSUE #142: Add type hints
https://docs.python.org/3/library/typing.html

jmorel:
Fixes #142 

* Adds mypy to the project
* New makefile target: `check-type`
* Adds mypy check to travis config
* Adds type hints to client.py
* Adds type hints to exception.py

I recommend reading https://realpython.com/python-type-checking/ as a good introduction to type hints.

My comments after doing this preliminary work:
* `typing.cast` is needed in quite a few places but feels very wrong as it looks like a cast occurring at runtime but isn't, it's only here for static type checking https://mypy.readthedocs.io/en/stable/casts.html#casts
* understanding how to a add type hints to a decorator was pretty hard and the result is a bit puzzling
* the `exceptions.py` would need a pretty big refactoring if we wanted to remove the `# type: ignore` as the issue is not with the return values but with the number of args and their types.
* exceptions raised by functions can't be documented through type hints, they have to be documented using docstrings.
* the generated doc isn't improved, only more verbose

But the larger issue I have is that these type hints don't solve our documentation issue in any way. We could create very well defined types for each type of asset, but it would be the same amount of work as using pydantic to serialize/deserialize dict into proper objects with none of the added benefits (runtime checks, manipulating objects instead of dicts).

I do feel type hints have a role to play though, but only once we have proper objects that are themselves well documented, and without the strict setup that I tried here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ sdist/
 .cache
 .coverage
 .pytest_cache/
+.mypy_cache/
 
 # Developer environments
 .idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,8 @@ deploy:
 script:
   - pip install flake8
   - flake8 substra
+  - pip install mypy
+  - make check-type
   - pip install -e .[test]
   - python setup.py test
   - python bin/generate_cli_documentation.py --output-path references/cli.md.tmp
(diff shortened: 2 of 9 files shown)


jmorel:
While this was interesting preliminary work, it's still too much work to maintain a strict policy on type hints with little added benefits since we only handle very basic types anyway.
I'm therefore closing the PR, but it could serve as reference for future efforts.

PR #149, opened 2020-03-24
Prevent aggregatuples with less than 2 inmodels

jmorel:
Companion PR: https://github.com/SubstraFoundation/substra-backend/pull/189

Prevent the creation (from the command line) of aggregatetuples with no in models or just 1 inmodel, since they don't make much sense (no aggregation to do).


jmorel:
Quoting @samlesu in the companion PR:
> After discussing with @camillemarini, it seems better to not add this limitation.
> There is no reason for restricting the aggregation of nothing or the aggregation of a single model.

PR #151, opened 2020-03-27
Prevent from path traversal when extracting archives

Kelvin-M:
Filenames should be validated when storing asset to prevent from path/directory traversal.
Companion PR :  https://github.com/SubstraFoundation/substra-backend/pull/192

### DIFF
--- a/substra/runner.py
+++ b/substra/runner.py
@@ -201,7 +201,22 @@ def compute_perf(pred_path, opener_file, fake_data_samples, data_path, docker_cl
     return perf['all']
 
 
-def raise_if_path_traversal(requested_paths, to_directory):
+class PathTraversalException(Exception):
+
+    def __init__(self, archive_path, issue_path):
+        self.msg = (
+            f'Path Traversal Error : archive {archive_path} contains '
+            f'{issue_path} which is not safe.'
+        )
+        super().__init__(self.msg)
+        self.archive_path = archive_path
+        self.issue_path = issue_path
+
+    def __repr__(self):
+        return self.msg
+
+
+def raise_if_path_traversal(requested_paths, to_directory, archive_path):
     # Inspired from https://stackoverflow.com/a/45188896
 
     # Get real path and ensure there is a suffix /
@@ -220,9 +235,9 @@ def raise_if_path_traversal(requested_paths, to_directory):
                                              safe_directory]) != safe_directory
 
         if is_traversal:
-            raise Exception(
-                f'Path Traversal Error : {requested_path} '
-                f'(real : {real_requested_path}) is not safe.'
+            raise PathTraversalException(
+                archive_path,
+                requested_path.replace(to_directory, '')
             )
 
 
@@ -233,7 +248,7 @@ def _extract_archive(archive_path, to_directory):
             # Check no path traversal
             filenames = [os.path.join(to_directory, filename)
                          for filename in zf.namelist()]
-            raise_if_path_traversal(filenames, to_directory)
+            raise_if_path_traversal(filenames, to_directory, archive_path)
 
             zf.extractall(to_directory)
 
@@ -243,7 +258,7 @@ def _extract_archive(archive_path, to_directory):
             # Check no path traversal
             filenames = [os.path.join(to_directory, filename)
                          for filename in tf.getnames()]
-            raise_if_path_traversal(filenames, to_directory)
+            raise_if_path_traversal(filenames, to_directory, archive_path)
 
             tf.extractall(to_directory)
     else:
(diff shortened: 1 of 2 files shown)


substra/runner.py
```diff
@@ -201,12 +201,48 @@ def compute_perf(pred_path, opener_file, fake_data_samples, data_path, docker_cl
...
+        path_traversal = os.path.commonprefix([real_requested_path,
+                                               safe_directory]) != safe_directory
+
+        if path_traversal:
+            raise Exception(
+                f'Path Traversal Error : {requested_path} '
+                f'(real : {real_requested_path}) is not safe.'
+            )
+
+
```
natct10: That's great!

substra/runner.py
```diff
@@ -201,12 +201,48 @@ def compute_perf(pred_path, opener_file, fake_data_samples, data_path, docker_cl
...
+        ''
+    )
+
+    if not isinstance(requested_paths, list):
+        raise TypeError(f'requested_paths argument should be a list not a {type(requested_paths)}')
+
+    for requested_path in requested_paths:
+        real_requested_path = os.path.realpath(requested_path)
+        path_traversal = os.path.commonprefix([real_requested_path,
+                                               safe_directory]) != safe_directory
```
AlexandrePicosson: Does it prevent from something like this :
```python
safe_directory='/tmp/test/'
real_requested_path='/tmp/test/../../bin/file'
```

Because `os.path.commonprefix()` would return `/tmp/test/` as it is the longest matching prefix which is equal to `safe_directory`. But in the end the requested path still do path traversal and unzip to an unsafe directory.

samlesu: It should be ok thanks to the usage of `realpath`:

```
In [1]: os.path.realpath('/tmp/test/../../bin/file')
Out[1]: '/private/bin/file'
```

Kelvin-M: @AlexandrePicosson, as Samuel said, 
`os.path.realpath` will resolve the `'../../'` to the realpath  (it also resolve simlinks)
So, in your example, it will raise an `Exception`. You can see the unit test in substra-backend to see that your example is handle :)

jmorel:
Tested and working fine! 

However since the run_local is a command line utility raising a standard exception produces an ugly output:

```sh
substra run-local ../../test.zip \
  --train-opener=assets/dataset/opener.py \
  --test-opener=assets/dataset/opener.py \
  --metrics=assets/objective/ \
  --fake-data-samples

  File "/Users/jeremy/.virtualenvs/substra/bin/substra", line 11, in <module>
    load_entry_point('substra', 'console_scripts', 'substra')()
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 1259, in invoke
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 1066, in invoke
  File "/Users/jeremy/.virtualenvs/substra/lib/python3.7/site-packages/click/core.py", line 610, in invoke
  File "/Users/jeremy/substra/substra/substra/cli/interface.py", line 1044, in run_local
    inmodels=inmodels)
  File "/Users/jeremy/substra/substra/substra/runner.py", line 342, in compute
    with extract_archive_if_needed(algo_path) as algo_path, \
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 112, in __enter__
    return next(self.gen)
  File "/Users/jeremy/substra/substra/substra/runner.py", line 259, in extract_archive_if_needed
    _extract_archive(path, tmp_dir)
  File "/Users/jeremy/substra/substra/substra/runner.py", line 236, in _extract_archive
    raise_if_path_traversal(filenames, to_directory)
  File "/Users/jeremy/substra/substra/substra/runner.py", line 224, in raise_if_path_traversal
    f'Path Traversal Error : {requested_path} '
Exception: Path Traversal Error : /var/folders/z9/tjd43wyd11j9h9tn51m84jjm0000gn/T/tmpew0z0llg/../substra.code-workspace (real : /private/var/folders/z9/tjd43wyd11j9h9tn51m84jjm0000gn/T/substra.code-workspace) is not safe.
```

Could you catch the raised exception in the run_local cli command and raise a click.Parameter error?
I'd have liked to have something like this displayed:

```
Usage: substra run-local [OPTIONS] ALGO
Try 'substra run-local --help' for help.

Error: Invalid value for 'ALGO': Path '../../test.zip' includes 1 file located outside the archive root folder:
    ../substra.code-workspace
```

(the message is very approximative of course, feel free to change it for whatever you think makes the most sense for the end user)

Kelvin-M:
@jmorel  updated :)

PR #154, opened 2020-04-07
Multi node examples

### LINKED ISSUE #25: Would be nice to have an example that use data submitted to two nodes
In the titanic example, data are submitted to one node only. It would be nice to have two scripts mimicking two actors that would add data in two nodes (two since in our local setup we use two nodes by default). This could be done in addition to the current example.
We could then in the compute plan example have an model that is created from data from two nodes.

jmorel:
Fixes #25

Converts all example to a 2-nodes setup, with train data samples are distributed evenly on each node.

I tried to be as minimal as possible in the changes.


jmorel:
Since we do not support simultaneous tokens for multiple config profiles at the moment (fixed by #139 ) we have to put `login()` calls into the scripts, effectively re-introducing the bug fixed by #175 .

jmorel:
Closing as this is way out of date.

PR #155, opened 2020-04-08
Torch example

samlesu:
This PR contains the following files:

- torch-reference.py: official torch example. This is useful to check the modifications required to run a torch example on substra
- assets/: assets implementation based on the above torch example
- substra-local.py: tooling to execute the substra workflow in a local python environment (it imports the previous assets and execute the training/testing tasks in the current python environment)
- substra-remote.py: single and main script to add all the assets to a Substra platform

The model is trained through two epochs on a set of 5 randomized samples and is tested on a test dataset. 

It has been translated to two traintuples (one per epoch) and a single testtuple.

### DIFF
--- a/examples/torch/README.md
+++ b/examples/torch/README.md
@@ -1,0 +1,11 @@
+# Torch example
+
+Adapt the CIFAR10 Classifier Torch tutorial to Substra.
+
+## Torch
+
+```
+python reference_tutorial.py
+```
+
+## Substra (requires a network up)
--- a/examples/torch/assets/algo/algo.py
+++ b/examples/torch/assets/algo/algo.py
@@ -1,0 +1,108 @@
+import logging
+
+import substratools as tools
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+
+
+logger = logging.getLogger(__name__)
+
+
+class Net(nn.Module):
+    def __init__(self):
+        super(Net, self).__init__()
+        self.conv1 = nn.Conv2d(3, 6, 5)
+        self.pool = nn.MaxPool2d(2, 2)
+        self.conv2 = nn.Conv2d(6, 16, 5)
+        self.fc1 = nn.Linear(16 * 5 * 5, 120)
+        self.fc2 = nn.Linear(120, 84)
+        self.fc3 = nn.Linear(84, 10)
+
+    def forward(self, x):
+        x = self.pool(F.relu(self.conv1(x)))
+        x = self.pool(F.relu(self.conv2(x)))
+        x = x.view(-1, 16 * 5 * 5)
+        x = F.relu(self.fc1(x))
+        x = F.relu(self.fc2(x))
+        x = self.fc3(x)
+        return x
+
+
+class TorchClassifier(tools.algo.Algo):
+
+    def _serialize_train_state(self, model, optimizer):
+        """Returns single object with model and optimizer states."""
+        return {
+            'model': model,
+            'optimizer': optimizer,
+        }
+
+    def _deserialize_train_state(self, state):
+        """Returns model and optimizer states from previous ouput model train task."""
+        return state['model'], state['optimizer']
+
+    def train(self, X, y, models, rank):
+        """Training task implementation."""
+        net = Net()
+        optimizer = optim.SGD(net.parameters(), lr=0.001, momentum=0.9)
+
+        if rank > 0:
+            model_state, optimizer_state = self._deserialize_train_state(
+                models[0])
+
+            net.load_state_dict(model_state)
+            optimizer.load_state_dict(optimizer_state)
+
+        criterion = nn.CrossEntropyLoss()
+
+        running_loss = 0.0
+        i = 0
+        for inputs, labels in zip(X, y):
+            # zero the parameter gradients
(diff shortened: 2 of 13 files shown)


examples/torch/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+Adapt the CIFAR10 Classifier Torch tutorial to Substra.
+
+## Torch
+
+```
+python reference_tutorial.py
+```
+
+## Substra (requires a network up)
```
jmorel, line 11: "C'est un peu court jeune homme"

examples/torch/assets/dataset/opener.py
```diff
@@ -0,0 +1,97 @@
...
+
+        transform = transforms.Compose(
+            [transforms.ToTensor(),
+             transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))])
+        shuffle = True
+        trainset = TorchDataset(data, labels, transform=transform)
+        trainloader = torch.utils.data.DataLoader(trainset, batch_size=4,
+                                                  shuffle=shuffle, num_workers=2)
+
+        # TODO this is very unefficient as it will load all the data in memory.
```
jmorel, line 68: We may want to look into this seriously before merging. Otherwise we'll be promoting inefficient code that could make Substra look bad.

examples/torch/assets/algo/algo.py
```diff
@@ -0,0 +1,108 @@
...
+        return self._serialize_train_state(net.state_dict(), optimizer.state_dict())
+
+    def predict(self, X, model):
+        """Predict task implementation."""
+        y_pred = []
+
+        model_state, _ = self._deserialize_train_state(model)
+
+        net = Net()
+        net.load_state_dict(model_state)
```
jmorel, line 90: I think this could be included in load_model so that the predict only does prediction.

jmorel: Aren't you missing `net.eval()` here ?

examples/torch/assets/algo/algo.py
```diff
@@ -0,0 +1,108 @@
...
+        """Training task implementation."""
+        net = Net()
+        optimizer = optim.SGD(net.parameters(), lr=0.001, momentum=0.9)
+
+        if rank > 0:
+            model_state, optimizer_state = self._deserialize_train_state(
+                models[0])
+
+            net.load_state_dict(model_state)
+            optimizer.load_state_dict(optimizer_state)
```
jmorel, line 57: I think this should be included in the load_model method for clarity.

examples/torch/substra-network.py
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    cp_spec = {
+        'traintuples': traintuples,
+        'testtuples': [testtuple_spec],
+    }
+    cp_id = ctx.client.add_compute_plan(cp_spec)['computePlanID']
+    print(f'CP added: id={cp_id}')
+
+    # TODO wait for the cp to be executed
```
jmorel, line 180: You could just print a command to help the user keep track of the compute plan.

examples/torch/substra-network.py
```diff
@@ -0,0 +1,185 @@
...
+    _client: substra.Client = None
+
+    @property
+    def client(self):
+        return self._client
+
+    def initialize(self):
+        self._client = substra.Client(profile_name=self.profile)
+        self._client.login()
+        return self
```
jmorel, line 40: I feel this is added complexity for code that is no different from have 2 global variables:
```
client = substra.Client(profile_name='node-1')
client.login()
assets_root = os.path.join(CURRENT_DIR, 'assets')
```
If we don't want to have these as global variables, they could be created in the __main__ block and passed as args to register(), no need for an object.

examples/torch/torch-reference.py
```diff
@@ -0,0 +1,110 @@
+"""Pytorch reference turorial script from (CIFAR10 classifier):
+https://pytorch.org/tutorials/beginner/blitz/cifar10_tutorial.html#sphx-glr-beginner-blitz-cifar10-tutorial-py
```
jmorel, line 2: We could just include the link to the tutorial in the README without including the whole scripts:
- less code to maintain
- data scientists interested in integration torch with substra will most probably already have knowledge of pytorch.

examples/torch/assets/algo/algo.py
jmorel: Also, shouldn't we add `model.train()` here?

examples/torch/assets/dataset/opener.py
samlesu: I think this is out of scope of this PR as it will involve more design that could lead to a change of the opener methods signature.
As you said, as we don't want to promote inefficient code, we could add on top of the TODO comment a warning in the README?

examples/torch/torch-reference.py
samlesu: I think we should keep the original file here.

The online tutorial may change in the future and we want to keep a version of it.
It will also allow to easily compare and validate the two versions (w and wo Substra).

IMO it's easier to compare two python scripts than comparing an online example and a script.

samlesu:
I'm closing this PR as it is not a priority now. It could be re-opened later.

PR #170, opened 2020-05-14
Fix ci (new release of flake8)

samlesu:
Release of flake8 3.8.0 is triggering a new warning in our code base. 
This PR is fixing it to keep using the latest version of flake8.


samlesu:
I think it's better to use the latest version of flake8.

PR #177, opened 2020-06-09
Local debugging

samlesu:
Progress:

- [x] Refactor the existing code to create a remote backend 
- [x] Add a local backend

Additional changes:

- the profile and configuration have been simplified in order to abstract properly the backends (the backends won't depend on it).
- use schemas to parse input arguments to simplify backend interface
- deprecate `Client.update_dataset` in favour of `Client.link_dataset_with_objective`

The current PR could be merge as if: it is not modifynig the existing public API.

### DIFF
--- a/substra/sdk/backends/local/backend.py
+++ b/substra/sdk/backends/local/backend.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import functools
 import shutil
 import uuid
 
@@ -222,16 +221,14 @@ class Local(base.BaseBackend):
         aggregatetuple_keys=None,
         testtuple_keys=None,
     ):
-        tuple_count = functools.reduce(
-            lambda x, y: x + (len(y) if y else 0),
-            [
+        tuple_count = sum([
+            (len(x) if x else 0) for x in [
                 traintuple_keys,
                 composite_traintuple_keys,
                 aggregatetuple_keys,
                 testtuple_keys,
-            ],
-            0,
-        )
+            ]
+        ])
         compute_plan = models.ComputePlan(
             compute_plan_id=uuid.uuid4().hex,
             status=models.Status.waiting.value,
--- a/substra/sdk/backends/local/backend.py
+++ b/substra/sdk/backends/local/backend.py
@@ -1,0 +1,527 @@
+# Copyright 2018 Owkin, inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import functools
+import shutil
+import uuid
+
+import substra
+from substra.sdk import schemas
+from substra.sdk.backends import base
+from substra.sdk.backends.local import models
+from substra.sdk.backends.local import db
+from substra.sdk.backends.local import fs
+from substra.sdk.backends.local import hasher
+from substra.sdk.backends.local import compute
+
+_BACKEND_ID = "local-backend"
+
+
+class Local(base.BaseBackend):
+    def __init__(self, *args, **kwargs):
+        # create a store to abstract the db
+        self._db = db.get()
+        self._worker = compute.Worker()
+
+    def get(self, asset_type, key):
+        return self._db.get(asset_type, key).to_response()
+
+    def list(self, asset_type, filters=None):
+        assets = self._db.list(asset_type)
+        return [a.to_response() for a in assets]
+
+    def __compute_permissions(self, permissions):
+        """Compute the permissions
+
+        If the permissions are private, the active node is
+        in the authorized ids.
+        """
+        if permissions.public:
+            permissions.authorized_ids = list()
+        elif not permissions.public and _BACKEND_ID not in permissions.authorized_ids:
+            permissions.authorized_ids.append(_BACKEND_ID)
+        return permissions
+
+    def _add_dataset(self, spec, exist_ok, spec_options):
+        permissions = self.__compute_permissions(spec.permissions)
+        asset = models.Dataset(
+            key=fs.hash_file(spec.data_opener),
+            pkhash=fs.hash_file(spec.data_opener),
+            owner=_BACKEND_ID,
+            name=spec.name,
+            objective_key=spec.objective_key if spec.objective_key else "",
+            permissions={
+                "process": {
+                    "public": permissions.public,
+                    "authorized_ids": permissions.authorized_ids,
+                },
+            },
+            type=spec.type,
+            train_data_sample_keys=[],
+            test_data_sample_keys=[],
+            data_opener=str(spec.data_opener),
+            description=str(spec.description),
+            metadata=spec.metadata if spec.metadata else dict(),
+        )
+        return self._db.add(asset, exist_ok)
+
+    def _add_data_sample(self, spec, exist_ok, spec_options):
+        assert len(spec.data_manager_keys) > 0
+        datasets = [
+            self._db.get(schemas.Type.Dataset, dataset_key)
+            for dataset_key in spec.data_manager_keys
+        ]
+
+        data_sample = models.DataSample(
+            key=fs.hash_directory(spec.path),
+            pkhash=fs.hash_directory(spec.path),
+            owner=_BACKEND_ID,
+            path=str(spec.path),
+            data_manager_keys=spec.data_manager_keys,
+            test_only=spec.test_only,
+        )
+        data_sample = self._db.add(data_sample, exist_ok)
+
+        # update dataset(s) accordingly
+        for dataset in datasets:
+            if spec.test_only:
+                samples_list = dataset.test_data_sample_keys
+            else:
+                samples_list = dataset.train_data_sample_keys
+            if data_sample.key not in samples_list:
+                samples_list.append(data_sample.key)
+
+        return data_sample
+
+    def _add_data_samples(self, spec, exist_ok, spec_options):
+        datasets = [
+            self._db.get(schemas.Type.Dataset, dataset_key)
+            for dataset_key in spec.data_manager_keys
+        ]
+
+        data_samples = [
+            models.DataSample(
+                key=fs.hash_directory(str(p)),
+                pkhash=fs.hash_directory(str(p)),
+                owner=_BACKEND_ID,
+                path=str(p),
+                data_manager_keys=spec.data_manager_keys,
+                test_only=spec.test_only,
+            )
+            for p in spec.paths
+        ]
+
+        data_samples = [self._db.add(a) for a in data_samples]
+
+        # update dataset(s) accordingly
+        for dataset in datasets:
+            if spec.test_only:
+                samples_list = dataset.test_data_sample_keys
+            else:
+                samples_list = dataset.train_data_sample_keys
+
+            for data_sample in data_samples:
+                if data_sample.key not in samples_list:
+                    samples_list.append(data_sample.key)
+
+        return data_samples
+
+    def _add_objective(self, spec, exist_ok, spec_options):
+        permissions = self.__compute_permissions(spec.permissions)
+        objective_key = fs.hash_file(spec.metrics)
+
+        # validate spec
+        test_dataset = None
+        if spec.test_data_manager_key:
+            dataset = self._db.get(schemas.Type.Dataset, spec.test_data_manager_key)
+            # validate test data samples
+            if spec.test_data_sample_keys is None:
+                spec.test_data_sample_keys = list()
+            for key in spec.test_data_sample_keys:
+                self._db.get(schemas.Type.DataSample, key)
+            test_dataset = {
+                "dataset_key": spec.test_data_manager_key,
+                "data_sample_keys": spec.test_data_sample_keys,
+            }
+            if not dataset.objective_key:
+                dataset.objective_key = objective_key
+            else:
+                raise substra.exceptions.InvalidRequest(
+                    "dataManager is already associated with a objective", 400
+                )
+
+        # create objective model instance
+        objective = models.Objective(
+            key=objective_key,
+            pkhash=objective_key,
+            name=spec.name,
+            owner=_BACKEND_ID,
+            test_dataset=test_dataset,
+            permissions={
+                "process": {
+                    "public": permissions.public,
+                    "authorized_ids": permissions.authorized_ids,
+                },
+            },
+            description=str(spec.description),
+            metrics=str(spec.metrics),
+            metadata=spec.metadata if spec.metadata else dict(),
+        )
+
+        # add objective to storage and update optionnally the associated dataset
+        objective = self._db.add(objective, exist_ok)
+
+        return objective
+
+    def __add_algo(self, model_class, spec, exist_ok, spec_options=None):
+        permissions = self.__compute_permissions(spec.permissions)
+        algo = model_class(
+            key=fs.hash_file(spec.file),
+            pkhash=fs.hash_file(spec.file),
+            name=spec.name,
+            owner=_BACKEND_ID,
+            permissions={
+                "process": {
+                    "public": permissions.public,
+                    "authorized_ids": permissions.authorized_ids,
+                },
+            },
+            file=str(spec.file),
+            description=str(spec.description),
+            metadata=spec.metadata if spec.metadata else dict(),
+        )
+        return self._db.add(algo, exist_ok)
+
+    def _add_algo(self, spec, exist_ok, spec_options=None):
+        return self.__add_algo(models.Algo, spec, exist_ok, spec_options=spec_options)
+
+    def _add_aggregate_algo(self, spec, exist_ok, spec_options=None):
+        return self.__add_algo(
+            models.AggregateAlgo, spec, exist_ok, spec_options=spec_options
+        )
+
+    def _add_composite_algo(self, spec, exist_ok, spec_options=None):
+        return self.__add_algo(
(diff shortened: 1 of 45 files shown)


Esadruhn:
- local backend: all client functions except 
  - `leaderboard`,  
  - `add_compute_plan`,  
  - `cancel_compute_plan`,  
  - `update_compute_plan`,  
  - `add_aggregatetuple`,  
  - `add_composite_traintuple`
- worker/spawner: the worker spawns a Docker container that the user can attach to during the execution (the container is removed once the execution is over), the execution is blocking
- client now accepts "schema" objects in addition to dictionaries
- [X] end-to-end test
- [X] documentation and example
- the list of permissions are computed for each object but there is no restriction on the user actions
- the dict returned to the user contains the `key` and `pkhash` as the remote but some objects are slightly different (e.g. return `metrics: str` instead of `metrics: {storageAddress: ..., ...}`)

substra/sdk/schemas.py
```diff
@@ -0,0 +1,267 @@
...
+            yield data, None
+
+
+class DatasetSpec(_Spec):
+    name: str
+    data_opener: str
+    type: str
+    description: str
+    permissions: Permissions
+    objective_key: Optional[str]
```
jmorel, line 126: Since the beginning of this PR, metadata have been added to all assets (but datasamples), they should be added in the specs as well.

substra/sdk/backends/local/db.py
```diff
@@ -0,0 +1,77 @@
...
+            logger.info(f"{type_} with key '{key}' has been created.")
+        return asset
+
+    def get(self, type_, key):
+        """Return asset."""
+        try:
+            return self._data[type_][key]
+        except KeyError:
+            logger.error(f"{type_} with key '{key}' not found.")
+            raise
```
jmorel, line 57: Why not raise exceptions.NotFound here?

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,493 @@
...
+            compute_plan_id = compute_plan.compute_plan_id
+            rank = (
+                len(compute_plan.traintuple_keys)
+                if compute_plan.traintuple_keys is not None
+                else 0 + len(compute_plan.composite_traintuple_keys)
+                if compute_plan.composite_traintuple_keys is not None
+                else 0 + len(compute_plan.aggregatetuple_keys)
+                if compute_plan.aggregatetuple_keys is not None
+                else 0
+            )
```
jmorel, line 296: I don't understand this code, could you explain it in comments?

My main issue is that for me, the rank of a traintuple is max of the in_models tuple ranks + 1 and I don't see that logic here. But maybe this way of calculating rank serves another purpose.

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,493 @@
...
+        traintuple = self._db.add(traintuple, exist_ok)
+        self._worker.schedule_traintuple(traintuple)
+        return traintuple
+
+    def _add_testtuple(self, spec, exist_ok, spec_options=None):
+
+        # validation
+        objective = self._db.get(schemas.Type.Objective, spec.objective_key)
+        traintuple = self._db.get(schemas.Type.Traintuple, spec.traintuple_key)
+        assert traintuple.out_model is not None
```
jmorel, line 348: I think we should raise a proper exception here for the user.

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,493 @@
...
+        if spec.data_manager_key is not None:
+            self._db.get(schemas.Type.Dataset, spec.data_manager_key)
+        if spec.test_data_sample_keys is not None:
+            [
+                self._db.get(schemas.Type.DataSample, key)
+                for key in spec.test_data_sample_keys
+            ]
```
jmorel, line 355: There should be a check that the data samples are linked to the data manager.

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,493 @@
...
+
+        # Hash creation
+        key_components = [_BACKEND_ID, spec.objective_key, spec.traintuple_key] + (
+            spec.test_data_sample_keys if spec.test_data_sample_keys is not None else []
+        )
+        if spec.data_manager_key is not None:
+            key_components += spec.data_manager_key
```
jmorel, line 362: Could we use the same style for the 2 ifs?

substra/sdk/backends/local/backend.py
Esadruhn: My logic was that when we add a traintuple, there may already be aggregate tuples, composite tuples and traintuples, so the rank is the number of pre-existing tuples + 1
If we say that the rank is number of in_models + 1, what if one of the in_models is already in the compute plan with a rank > 0 ?

Can there be several tuples with the same rank in the compute plan ?

Esadruhn: my bad, ok for max of the in_models tuple ranks + 1

substra/sdk/schemas.py
Esadruhn: Should it be added to 'models' as well ?
The type is Optional[dict] I suppose ?

substra/sdk/backends/local/backend.py
Esadruhn: my bad, updated !

substra/sdk/schemas.py
jmorel: In specs, the type is `Optional[Dict[str,str]]` and in models it is `Dict[str,str]`: optional when submitted, but always a dict when returned by the backend (an empty dict in case of no value submitted).

substra/sdk/backends/local/backend.py
```diff
@@ -285,15 +285,14 @@ def _add_traintuple(self, spec, exist_ok, spec_options=None):
...
-                if compute_plan.composite_traintuple_keys is not None
-                else 0 + len(compute_plan.aggregatetuple_keys)
-                if compute_plan.aggregatetuple_keys is not None
-                else 0
+            rank = 1 + max(
+                [
+                    in_traintuple.rank
+                    for in_traintuple in in_traintuples
+                    if in_traintuple.compute_plan_id == compute_plan_id
+                ]
```
jmorel, line 293: Beware, if there are no in_traintuples, then the rank is 0 :)

substra/sdk/backends/local/backend.py
```diff
@@ -349,15 +348,18 @@ def _add_testtuple(self, spec, exist_ok, spec_options=None):
         if spec.data_manager_key is not None:
             self._db.get(schemas.Type.Dataset, spec.data_manager_key)
         if spec.test_data_sample_keys is not None:
-            [
-                self._db.get(schemas.Type.DataSample, key)
+            same_data_manager = all([
+                self._db.get(schemas.Type.DataSample, key).data_manager_key == spec.data_manager_key
```
jmorel, line 352: I believe datasamples do not have a single data_manager_key but many (they can be attached to multiple datasets).

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,523 @@
...
+        return self.__add_algo(
+            models.CompositeAlgo, spec, exist_ok, spec_options=spec_options
+        )
+
+    def __intersect_permissions(self, aggregate_set, element):
+        if not element.permissions.process.public:
+            if aggregate_set is None:
+                aggregate_set = set(element.permissions.process.authorized_ids)
+            else:
+                aggregate_set = aggregate_set & set(
```
jmorel, line 218: Could you use the `intersection` method here instead of `&` ? I find it much more explicit and the doc stresses that it's more readable.

jmorel: Also, this is a static method, maybe it could be moved to an independant function instead of a method?

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,523 @@
...
+        tuple_count = functools.reduce(
+            lambda x, y: x + (len(y) if y else 0),
+            [
+                traintuple_keys,
+                composite_traintuple_keys,
+                aggregatetuple_keys,
+                testtuple_keys,
+            ],
+            0,
+        )
```
jmorel, line 239: For the sake of clarity, I'd rather we use `tuple_count = len(traintuple_keys) + len(composite_traintuple_keys) + len(aggregatetuple_keys) + len(testtuple_keys)`.

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,523 @@
...
+        key = hasher.Hasher(values=key_components).compute()
+
+        # permissions
+        with_permissions = [algo, data_manager] + in_traintuples
+
+        authorized_ids = functools.reduce(
+            self.__intersect_permissions, with_permissions, None
+        )
+        public = authorized_ids is None
```
jmorel, line 291: What happens if there was an error and for example the algo is private to only A and the dataset is private to only B? Normally the traintuple won't be accepted, but I read the code as "the traintuple is accepted and the model permission will be public".

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,523 @@
...
+                [
+                    spec.data_manager_key
+                    in self._db.get(schemas.Type.DataSample, key).data_manager_keys
+                    for key in spec.test_data_sample_keys
+                ]
+            )
+            if not same_data_manager:
+                raise substra.exceptions.InvalidRequest(
+                    "dataSample do not belong to the same dataManager", 400
+                )
```
jmorel, line 384: It looks like this code is duplicated in this module, it could be moved to a separate function.

substra/sdk/schemas.py
```diff
@@ -0,0 +1,282 @@
...
+    test_data_sample_keys: Optional[List[str]]
+    metadata: Optional[Dict[str, str]]
+
+
+class _BaseComputePlanSpec(_Spec, abc.ABC):
+    traintuples: Optional[List[ComputePlanTraintupleSpec]]
+    composite_traintuples: Optional[List[ComputePlanCompositeTraintupleSpec]]
+    aggregatetuples: Optional[List[ComputePlanAggregatetupleSpec]]
+    testtuples: Optional[List[ComputePlanTesttupleSpec]]
+    metadata: Optional[Dict[str, str]]
```
jmorel, line 271: metadata should be in COmputePlanSpec, they are not accepted during update.

tests/sdk/conftest.py
```diff
@@ -17,11 +17,18 @@
...
+
 @pytest.fixture
 def client(tmpdir):
-    config_path = tmpdir / "substra.cfg"
-    c = substra.Client(config_path=str(config_path))
-    c.add_profile('test', url="http://foo.io")
+    # config_path = tmpdir / "substra.cfg"
+    c = substra.Client(url="http://foo.io")
+    # substra.Client.load_profile(config_path=str(config_path))
+    # c.add_profile("test", url="http://foo.io")
```
jmorel, line 31: extra comments could be removed.

tests/sdk/test_config.py
```diff
@@ -74,21 +74,19 @@ def test_load_profile_fail(tmpdir):
 
 
 def test_login_without_profile(tmpdir):
-    client = substra.Client()
-
     with pytest.raises(substra.exceptions.SDKException):
-        client.login('foo', 'bar')
+        substra.Client()
```
jmorel, line 78: The name of the test is misleading.

substra/sdk/client.py
```diff
@@ -44,132 +47,95 @@ def wrapper(*args, **kwargs):
...
+    """Create the client.
+
+    Set the `backend` to 'local' to use the local debugging.
+    One client corresponds to one profile.
+    """
+
+    def __init__(
+        self,
+        url: typing.Optional[str] = None,
+        user_path: typing.Optional[str] = None,
```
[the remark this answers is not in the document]
Esadruhn: in this case the token is not Optional I guess ?

substra/sdk/client.py
```diff
@@ -44,132 +47,95 @@ def wrapper(*args, **kwargs):
...
+        return self._token
+
+    @staticmethod
+    def _get_spec(asset_type, data):
+        if isinstance(data, asset_type):
+            return data
+        return asset_type(**data)
+
+    @classmethod
+    def load_profile(
```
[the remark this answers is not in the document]
Esadruhn: perfect ! :)

substra/sdk/config.py
```diff
@@ -96,27 +96,28 @@ def _load_profile(path, name):
...
         return config[name]
     except KeyError:
         raise ProfileNotFoundError(name)
 
 
-class Manager():
+class Manager:
     def __init__(self, path=DEFAULT_PATH):
+        if not path:
+            path = DEFAULT_PATH
```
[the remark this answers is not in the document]
Esadruhn: Well it is in the doc and CLI description that the default path is ~/.substra. Maybe we can keep it but add a logger.info("using the default configuration file {DEFAULT_PATH}") ?

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,523 @@
...
+        data_sample = self._db.add(data_sample, exist_ok)
+
+        # update dataset(s) accordingly
+        for dataset in datasets:
+            if spec.test_only:
+                samples_list = dataset.test_data_sample_keys
+            else:
+                samples_list = dataset.train_data_sample_keys
+            if data_sample.key not in samples_list:
+                samples_list.append(data_sample.key)
```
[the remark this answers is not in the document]
Esadruhn: datasets are the datasets whose keys are in spec.data_manager_keys, see above:

```python
datasets = [
            self._db.get(schemas.Type.Dataset, dataset_key)
            for dataset_key in spec.data_manager_keys
        ]
```

substra/sdk/backends/local/backend.py
```diff
@@ -0,0 +1,523 @@
...
+            },
+            description=str(spec.description),
+            metrics=str(spec.metrics),
+            metadata=spec.metadata if spec.metadata else dict(),
+        )
+
+        # add objective to storage and update optionnally the associated dataset
+        objective = self._db.add(objective, exist_ok)
+        if spec.test_data_manager_key:
+            dataset.objective_key = objective.key
```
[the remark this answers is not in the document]
Esadruhn: good catch !

substra/sdk/backends/local/backend.py
Esadruhn: I agree but any of these can either be a list or None, I can do

```python
tuple_count = 0
if traintuple_keys is not None:
    tuple_count += len(traintuple_keys)
if ....
```

but I would like to avoid

```python
tuple_count = (len(traintuple_keys) if traintuple_keys is not None else 0) + ...
```

substra/sdk/backends/local/backend.py
Esadruhn: Indeed there is a problem in the case "private and no permission"

substra/sdk/client.py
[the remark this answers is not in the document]
jmorel: I think it should be optional so that we can use login later on.

substra/sdk/config.py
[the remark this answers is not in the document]
jmorel: The issue is that the method already has `path=DEFAULT_PATH` as argument, so the only way we would enter the if is if someone write `ConfigManager(None)`. I'm just not at ease with both way to specify a default value, one or the other would be better I think. But this is just my point of view and not an absolute.

I'd like to have both ConfigManager and UserManager use the same pattern for the defualt value though.

substra/sdk/backends/local/backend.py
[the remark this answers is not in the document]
jmorel: Oh my bad then, thanks!

substra/sdk/client.py
```diff
@@ -44,132 +47,92 @@ def wrapper(*args, **kwargs):
...
+        if isinstance(data, asset_type):
+            return data
+        return asset_type(**data)
+
+    @classmethod
+    def from_config_file(
+        cls,
+        profile_name: str,
+        config_path: typing.Union[str, pathlib.Path, None] = None,
+        user_path: str = None,
```
jmorel, line 119: The type should be the same as config_path's.

substra/sdk/client.py
```diff
@@ -44,132 +47,92 @@ def wrapper(*args, **kwargs):
...
+    @staticmethod
+    def _get_spec(asset_type, data):
+        if isinstance(data, asset_type):
+            return data
+        return asset_type(**data)
+
+    @classmethod
+    def from_config_file(
+        cls,
+        profile_name: str,
```
jmorel, line 117: Weirdly enough, I think profile_name should be optional as well, since there is a default value.

jmorel:
The CLi is struggling to load tokens:

```
[substra] ~/substra/substra-tests (master)
➜ substra config http://substra-backend.node-1.com
[substra] ~/substra/substra-tests (master)
➜ substra login --username node-1 --password 'p@$swr0d44'
Token: 033521360c304c55785329efb699ec39bceae7e8
[substra] ~/substra/substra-tests (master)
➜ substra list dataset
Requests error status 401: {"detail":"Invalid token."}
Error: Request failed: AuthenticationError: 401 Client Error: Unauthorized for url: http://substra-backend.node-1.com/data_manager/
✗✗✗
```

substra/sdk/config.py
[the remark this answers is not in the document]
Esadruhn: ok, I didn't pay attention that it was already the default value in the signature, I change it to have 
```
def __init__(self, path=DEFAULT_PATH):
        self.path = path
```
in both cases

substra/sdk/backends/local/backend.py
jmorel: Got it. What do you think about using sum() instead of reduce then (with a lambda to handle the None case)?

substra/sdk/utils.py
```diff
@@ -36,6 +36,7 @@ def path_leaf(path):
 
 @contextlib.contextmanager
 def extract_files(data, file_attributes):
+    # FIXME should be move to the backend module
```
jmorel, line 39: Could you implement these 2 fixme's? They seem simple enough to be included in this PR.

Esadruhn: extract_files and extract_data_sample_files are used in schema.py, which is in substra/sdk so it should stay there, I can remove the FIXME

jmorel: That's good for me :)

PR #190, opened 2020-06-22
Add metadata in get command's output

maeldebon:
This PR aims to add the metadata field in the get command's output.

### DIFF
--- a/substra/cli/printers.py
+++ b/substra/cli/printers.py
@@ -53,6 +53,15 @@ class Field:
                         print(f'{padding}- {v}')
             else:
                 print(f'{name}None')
+        elif isinstance(value, dict):
+            if len(value):
+                padding = ' ' * field_length
+                print(f'{name}')
+                for k, v in value.items():
+                    print(f'{padding}- {k}: {v}')
+            else:
+                name = self.name.upper().ljust(field_length)
+                print(f'{name}None')
         else:
             print(f'{name}{value}')
 
@@ -77,22 +86,6 @@ class PermissionField(Field):
             print(f'{name}Processable by:')
             for v in value:
                 print(f'{padding}- {v}')
-        else:
-            super().print_details(item, field_length, expand)
-
-
-class MetadataField(Field):
-    def print_details(self, item, field_length, expand):
-        metadata = find_dict_composite_key_value(item, 'metadata')
-        if isinstance(metadata, dict) and len(metadata):
-            name = self.name.upper().ljust(field_length)
-            padding = ' ' * field_length
-            print(f'{name}Metadata for this asset:')
-            for k, v in metadata.items():
-                print(f'{padding}- {k}:{v}')
-        elif isinstance(metadata, dict):
-            name = self.name.upper().ljust(field_length)
-            print(f'{name}None')
         else:
             super().print_details(item, field_length, expand)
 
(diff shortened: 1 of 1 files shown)


jmorel:
Could you provide an output example? Looking at the code I fear it just prints the dict as is with no formatting.

maeldebon:
> Could you provide an output example? Looking at the code I fear it just prints the dict as is with no formatting. Also it looks like it's printed in the list which could make the list command unusable as soon as there are metadata that are longer than `foo: bar`.

Indeed, it actually returns the dict. I'll switch to the same format as permissions.

jmorel:
Could you add an example of the output ?

maeldebon:
> Could you add an example of the output ?

Yes!
```
OBJECTIVE KEY
TYPE                        Test
TRAIN DATA SAMPLE KEYS      None
TEST DATA SAMPLE KEYS       None
OWNER                       MyOrg1MSP
METADATA                    Metadata for this asset:
                            - bar:foo
                            - foo:bar
PERMISSIONS                 Processable by anyone
```

jmorel:
Thanks for the output! A couple of remarks though:
* since we already have `METADATA` at the beginning of the line, `Metadata for this asset` is a bit redundant.
* in typography, there is always a space after a colon (`:`)

My last point is that if you remove the `Metadata for this asset`, there is nothing specific about the MetadataField anymore, apart from the fact that it handles dicts. So you could update the Field class so that it handles strings, lists (it does already the two) and dicts by refactoring the code you just wrote. This would also simplify things for future dict values.

substra/cli/printers.py
```diff
@@ -266,6 +275,7 @@ class ComputePlanPrinter(AssetPrinter):
         ProgressField('Progress', 'doneCount', 'tupleCount'),
         Field('Status', 'status'),
         Field('Tag', 'tag'),
+        Field('Metadata', 'metadata'),
```
jmorel, line 278: It looks like there are a few metadata fields in the "list_fields" sections, they shouldn't be here as they will break the output (they are too long to be printed in a table).

substra/cli/printers.py
```diff
@@ -53,6 +53,15 @@ def print_details(self, item, field_length, expand):
...
                 print(f'{name}None')
+        elif isinstance(value, dict):
+            if len(value):
+                padding = ' ' * field_length
+                print(f'{name}')
+                for k, v in value.items():
+                    print(f'{padding}- {k}: {v}')
+            else:
+                name = self.name.upper().ljust(field_length)
+                print(f'{name}None')
```
jmorel, line 64: Here is the output of this code:
```
METADATA
                            - foo: bar
```

The -foo:bar should be on the same line as METADATA here, there's no reason to have an extraline.

If you look at the code on L45-46, this is achieved by adding `end=''` to the first print, thus not printing the carriage return after `name`. Then the first item of the list is not printed with padding.

Since printing list and dicts is almost the same, you could transform the dict value as a list value and then let the existing code handle the printing.

For example:

```python
if isinstance(value, dict):
    dict = [f{k}: {v}' for k, v in value.items()]
```
```

**Comment by maeldebon** [comment]
Here's the output:
```
KEY                         0fb5b8d1a97174a09bfc8a34b5884d1d0d0ad5127ef082155bc724e4fc60714a
NAME                        24efe1ff14c24a09bd21832475ec380c_test_asset_with_metadata[metadata0-metadat...24efe1ff14c24a09bd21
OBJECTIVE KEY
TYPE                        Test
TRAIN DATA SAMPLE KEYS      None
TEST DATA SAMPLE KEYS       None
OWNER                       MyOrg1MSP
METADATA                    - foo: bar
PERMISSIONS                 Processable by anyone

Download this data manager's opener:
	substra download dataset 0fb5b8d1a97174a09bfc8a34b5884d1d0d0ad5127ef082155bc724e4fc60714a

Display this dataset's description:
	substra describe dataset 0fb5b8d1a97174a09bfc8a34b5884d1d0d0ad5127ef082155bc724e4fc60714a
```

PR #194, opened 2020-07-07
Add n_samples argument for fake data

Esadruhn:
https://github.com/SubstraFoundation/substra-tools/pull/38

### DIFF
--- a/examples/titanic/assets/dataset/opener.py
+++ b/examples/titanic/assets/dataset/opener.py
@@ -24,11 +24,11 @@ class TitanicOpener(tools.Opener):
         return pd.read_csv(path)
 
     def fake_X(self, n_samples=None):
-        data = self._fake_data()
+        data = self._fake_data(n_samples)
         return self._get_X(data)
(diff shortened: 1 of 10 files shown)


examples/titanic/assets/dataset/opener.py
```diff
@@ -23,11 +23,11 @@ def save_predictions(self, y_pred, path):
...
         return pd.read_csv(path)
 
-    def fake_X(self):
+    def fake_X(self, n_samples=None):
         data = self._fake_data()
         return self._get_X(data)
 
-    def fake_y(self):
+    def fake_y(self, n_samples=None):
         data = self._fake_data()
```
jmorel, line 31: According to the code, you should pass `n_samples` to the` _fake_data method`.

jmorel:
I just realized that this cannot work until the substra-tools docker images are updated. We'll have to make a new release over there, wait for the images to be built and then update the dockerfiles in the examples to point to the new release.

PR #195, opened 2020-07-07
DEPRECATED - Hybrid mode - local debugging on remote dataset (using fake_X, fake_y methods from the opener)

Esadruhn:
## **DEPRECATED**

This PR is no longer viable following the changes made to the "hybrid debugging" feature. It stays as a reference for now, will be deleted when the feature is integrated.

## Description

1. [X] Add fake data execution for local backend
2. [X] Once the PR https://github.com/SubstraFoundation/substra/pull/194 is merged, add the "n_samples" arg for fake data local debugging
3. [X] Add an example of hybrid debugging:
    1. Download the dataset from a remote client
    2. Add the dataset to the local client
    3. Run the local debugging with the 'fake data' option

## Main changes:

- add fake_data, n_fake_samples to the spec options when adding a train or test tuple with the sdk
- fix the add_traintuple and add_testtuple functions: do not schedules the tuples again if they existed in the database
- add tests and update the factory to be more thorough
- [X] Update for the composite  traintuples (not the aggretate tuples since they do not use data)
- [X] Update for the compute plan
- [X] Generate a different (random) hash for traintuples and testtuples that run on fake data (the hash should not use 'data_samples_keys')
- [X] error if the user gives tries using fake data with the remote backend


substra/sdk/client.py
```diff
@@ -376,12 +376,20 @@ def add_traintuple(self, data, exist_ok=False):
...
-        return self._backend.add(spec, exist_ok=exist_ok)
+        spec_options = {
+            'fake_data': fake_data,
+            'n_fake_samples': n_fake_samples
+        }
+        return self._backend.add(spec, exist_ok=exist_ok, spec_options=spec_options)
 
     @logit
-    def add_aggregatetuple(self, data, exist_ok=False):
+    def add_aggregatetuple(self, data, exist_ok=False, fake_data=False, n_fake_samples=None):
```
jmorel, line 392: There are no data samples for aggregate tasks, I think you can remove this part.

substra/sdk/client.py
```diff
@@ -376,12 +376,20 @@ def add_traintuple(self, data, exist_ok=False):
 
         If `exist_ok` is true, `AlreadyExists` exceptions will be ignored and the
         existing asset will be returned.
+
+        If `fake_data` is true and the backend is local, the traintuple runs on fake data.
+        If `fake_data` is true and `n_fake_samples` is set, the traintuple
```
jmorel, line 381: What do you think about raising an exception if the backend is remote?

substra/sdk/backends/local/backend.py
```diff
@@ -399,7 +399,11 @@ def _add_traintuple(self, spec, exist_ok, spec_options=None):
         )
 
         traintuple = self._db.add(traintuple, exist_ok)
-        self._worker.schedule_traintuple(traintuple)
+        fake_data = spec_options is not None and spec_options.get("fake_data", False)
+        n_fake_samples = spec_options.get("n_fake_samples")
```
jmorel, line 403: Shouldn't you test that spec options is defined here as well?

jmorel: Also, I think you're missing fake data sample handling for composite traintuples.

examples/local_debugging/README.md
```diff
@@ -1,6 +1,14 @@
...
-This example shows how to run the [Titanic example](../titanic/README.md) locally to debug your code.
+- [Local debugging](#local-debugging)
+  - [Prerequisites](#prerequisites)
+  - [Data and script preparation](#data-and-script-preparation)
+  - [Run and debug our pipeline](#run-and-debug-our-pipeline)
+  - [Difference with the Titanic example](#difference-with-the-titanic-example)
+- [Hybrid debugging](#hybrid-debugging)
+
+This example shows how to run the [Titanic example](../titanic/README.md) locally to debug your code. There are two possibilities:
+either run it completely locally or use the fake data from the dataset from the deployed Substra platform with a local client.
```
jmorel, line 11: The difference between the two possibilities may be obscure to most users, what do you think of:

```md
There are two possibilities as to which data to use for debugging:
* if you have data samples available on disk: use those
* if you don't: use fake data samples containing random data generated on the fly by the opener
```

substra/sdk/client.py
Esadruhn: Indeed, remote backend and fake_data = True should raise an exception.

examples/local_debugging/README.md
Esadruhn: Better ! I would only change
> use fake data samples containing random data generated on the fly by the opener

to

> use random data generated on the fly by the opener

because there are no data samples created.

jmorel: Very good point!

jmorel:
Regarding the move of fake_data / n_fake_data_Samples in the tuple specs, I'd be more at ease if this was only accepted for the local backend so that users do not expect a fake data tuple to work with a remote backend.

Esadruhn:
> Regarding the move of fake_data / n_fake_data_Samples in the tuple specs, I'd be more at ease if this was only accepted for the local backend so that users do not expect a fake data tuple to work with a remote backend.

We have several options for this:
1. a different "schemas" for the local and remote backend
2.  the local backend one uses schemas that inherit from the remote one (so it can add fields but not remove any)
3. or a test on the 'add' function on the remote backend that raises an error if the asset has fake data (current solution)

The advantage of the test is that the user can still use the same objects for local and remote, but it requires an additional test and a specific one for the compute plan, a big change for a small thing.

Esadruhn:
See PR https://github.com/SubstraFoundation/substra-tests/pull/121 for the tests.

PR #196, opened 2020-07-07
fix(sdk): client - set default value if manager gives None for path

Esadruhn:
Fix the bug when trying to run the Titanic example (gave None for config path instead of DEFAULT_CONFIG if the user did not give the argument)

```
  File "./scripts/add_dataset_objective.py", line 49, in <module>
    client = substra.Client.from_config_file(profile_name="node-1")
  File "/media/chrisalexandre/datalinux/dev/substra/substra/examples/titanic/.venv/lib/python3.6/site-packages/substra/sdk/client.py", line 125, in from_config_file
    profile = cfg_manager.from_config_file(name=profile_name)
  File "/media/chrisalexandre/datalinux/dev/substra/substra/examples/titanic/.venv/lib/python3.6/site-packages/substra/sdk/config.py", line 128, in from_config_file
    return _from_config_file(self.path, name)
  File "/media/chrisalexandre/datalinux/dev/substra/substra/examples/titanic/.venv/lib/python3.6/site-packages/substra/sdk/config.py", line 100, in _from_config_file
    config = _read_config(path)
  File "/media/chrisalexandre/datalinux/dev/substra/substra/examples/titanic/.venv/lib/python3.6/site-packages/substra/sdk/config.py", line 46, in _read_config
    if not os.path.exists(path):
  File "/usr/lib/python3.6/genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
```

### DIFF
--- a/substra/sdk/config.py
+++ b/substra/sdk/config.py
@@ -109,13 +109,19 @@ def _from_config_file(path, name):
 
 class Manager:
     def __init__(self, path=DEFAULT_PATH):
-        self.path = path
+        self.path = path or DEFAULT_PATH
 
     def add_profile(
         self, name, url=DEFAULT_URL, version=DEFAULT_VERSION, insecure=False
     ):
         """Add profile to config file on disk."""
-        config = _add_profile(self.path, name, url, version=version, insecure=insecure,)
+        config = _add_profile(
+            self.path,
+            name,
+            url or DEFAULT_URL,
+            version=version or DEFAULT_VERSION,
+            insecure=insecure,
+        )
         return config[name]
 
     def from_config_file(self, name=DEFAULT_PROFILE_NAME):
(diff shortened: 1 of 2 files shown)


jmorel:
I went through the code again and I think we may have started with the wrong idea when we wrote the `Manager` classes. The current classes all have their only init arg, `path`, as optional. And then we have lots of code to handle the case when there was no value provided.  

Since this class is for internal use only, we could make is so that it doesn't have any default and it's the role of the `Client.from_config_file` method to provide a correct value. I feel this could simplify the code quite a bit.

Esadruhn:
Indeed it would be better ! The code is hard to read, having only one config class to manage the profiles and tokens would be more straight-forward

In the meantime, do we merge the fix or is it OK to wait until the refacto ?  
It breaks the examples but the workaround is easy to implement (see the PR https://github.com/SubstraFoundation/substra/pull/198), I don't have enough visibility on this to know how impactful it is

jmorel:
Since I had a couple of hours to spare I wrote the refactoring: #199 let me know what you think.

PR #197, opened 2020-07-08
Local backend - add support for composite and aggregate traintuples

Esadruhn:
## Main changes
- Substra sdk, local backend: add the functions `add_composite_traintuple`, `add_aggregate_traintuple`
- Substra sdk, schemas: update the pydantic typing: pathlib.Path accepts Path and str
- added checks on the metadata, this is not directly related to composite/aggregate but they need the checks too.
  - each key in metadata can be of length 50 max
  - each value in metadata can be of length 100 max and cannot be empty

- [ ] Hybrid debugging: run these on fake data (wait for https://github.com/SubstraFoundation/substra/pull/195)

### DIFF
--- a/substra/sdk/backends/local/backend.py
+++ b/substra/sdk/backends/local/backend.py
@@ -15,7 +15,7 @@ import shutil
 import uuid
 
 import substra
-from substra.sdk import schemas
+from substra.sdk import schemas, exceptions
 from substra.sdk.backends import base
 from substra.sdk.backends.local import models
 from substra.sdk.backends.local import db
@@ -39,7 +39,8 @@ class Local(base.BaseBackend):
         assets = self._db.list(asset_type)
         return [a.to_response() for a in assets]
 
-    def __compute_permissions(self, permissions):
+    @staticmethod
+    def __compute_permissions(permissions):
         """Compute the permissions
 
         If the permissions are private, the active node is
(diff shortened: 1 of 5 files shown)


Esadruhn:
I updated the composite_traintuple creation so that out_head_model and out_trunk_model are created at the beginning with their permissions. This is because I realised that in substra-tests it expected failed composite traintuples to have out_head_models and out_trunk_models attributes.

More precisely:
- the client is shared between all tests (scope=session)
- the test test_data_samples_order.py::test_composite_traintuple_data_samples_relative_order fails as it is expeced to do, so there is a composite_traintuple in the database without out_head_model nor out_trunk_model
- some of the following tests call `client.list_compiosite_traintuple`, but in `substra-tests/assets.py/CompositeTraintuple`, `out_head_model` and `out_trunk_model` cannot be None. So those tests fail.

substra/sdk/backends/local/backend.py
```diff
@@ -244,6 +245,56 @@ def __add_compute_plan(
...
+                )
+
+            # Add to the compute plan
+            # TODO better way of doing this ?
+            if spec.__class__.type_ == schemas.TraintupleSpec:
+                compute_plan_attr_name = "traintuple_keys"
+            elif spec.__class__.type_ == schemas.CompositeTraintupleSpec:
+                compute_plan_attr_name = "composite_traintuple_keys"
+            elif spec.__class__.type_ == schemas.AggregatetupleSpec:
+                compute_plan_attr_name = "aggregatetuple_keys"
```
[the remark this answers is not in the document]
Esadruhn: yes, good idea

substra/sdk/backends/local/backend.py
```diff
@@ -369,10 +391,21 @@ def _add_traintuple(self, spec, exist_ok, spec_options=None):
...
+            traintuple = self._db.get(schemas.Type.Traintuple, spec.traintuple_key)
+            traintuple_type = schemas.Type.Traintuple
+        except exceptions.NotFound:
+            try:
+                traintuple = self._db.get(schemas.Type.CompositeTraintuple, spec.traintuple_key)
+                traintuple_type = schemas.Type.CompositeTraintuple
+            except exceptions.NotFound:
+                traintuple = self._db.get(schemas.Type.Aggregatetuple, spec.traintuple_key)
+                traintuple_type = schemas.Type.Aggregatetuple
+
```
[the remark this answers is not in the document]
Esadruhn: Thank you, it is better !

substra/sdk/schemas.py
```diff
@@ -191,6 +194,8 @@ class TraintupleSpec(_Spec):
     rank: Optional[int]
     metadata: Optional[Dict[str, str]]
 
+    compute_plan_attr_name: str = "traintuple_keys"
```
jmorel, line 197: What do you think of making these classvars (same as `type_`) ? This way it won't show up in `spec.dict()`.

substra/sdk/backends/local/compute/worker.py
```diff
@@ -72,30 +125,103 @@ def _context(self, key):
             # delete tuple working directory
             shutil.rmtree(tmp_dir, ignore_errors=True)
 
-    def schedule_traintuple(self, tuple_):
+    def schedule_composite_traintuple(self, tuple_):
```
jmorel, line 128: The schedule_composite_traintuple and schedule_traintuple are pretty similar and schedule_traintuple already handles both train and aggregatetuples. Do you think this could be refactored to have a single schedule_traintuple method that would handle all types?

substra/sdk/backends/local/compute/worker.py
```diff
@@ -72,30 +125,103 @@ def _context(self, key):
...
+    def schedule_traintuple(self, tuple_):
+        """Schedules a ML task (blocking)."""
+        with self._context(tuple_.key) as tuple_dir:
+            tuple_.status = models.Status.doing
+
+            # fetch dependencies
+            if isinstance(tuple_, models.Traintuple):
+                algo = self._db.get(schemas.Type.Algo, tuple_.algo_key)
+            else:
+                algo = self._db.get(schemas.Type.AggregateAlgo, tuple_.algo_key)
```
jmorel, line 213: You could use the same strategy used for the `type_` class attribute and add an `algo_type` attribute so that you can write `algo = self._db.get(tuple_.__class__.algo_type, tuple_.algo_key)`

Esadruhn: Good, the only thing that is a bit weird is that here `tuple_` is of type `models.Traintuple` but in the backend, `tuple_` is of class `schemas.TraintupleSpec`, so I added the `algo_type` to the models, and the `compute_plan_attr_name` to the schemas.

substra/sdk/backends/local/compute/worker.py
Esadruhn: I can let's see how it looks, 
I am not entirely convinced because the 'out_models' attribute has a different name, the commands and volumes are different too, but the steps are indeed the same.

Esadruhn: Done, there are "ifs" at each step but it is actually easy to read I think

PR #198, opened 2020-07-08
Debug titanic

chrisalexandrepena:


jmorel:
Thanks @chrisalexandrepena for this demonstration of the workaround. We're fixing the underlying bug (see #196 and #199) so I'd prefer we do not merge the workaround but keep the PR open until the bug is fixed.

jmorel:
Hello @chrisalexandrepena, a fix to this issue has been merged (in #199), could you check that this workaround is not necessary anymore and close your PR if it isn't?

jmorel:
Got confirmation from @chrisalexandrepena that it's indeed fixed and the workaround is not needed.

PR #199, opened 2020-07-08
Support multiple tokens (one per profile)

jmorel:
Adds support for one token per profile (instead of a single token at a time). From a user standpoint, this means needing to login less often.

Main changes:
* the tokens aren't stored in `~/.substra-user` anymore, but in `~/.substra-tokens`
* the Client API changes, the `user_path` parameter is renamed as `tokens_path`
* the ConfigManager and UserManager now inherit from a common _ProfileManager

Small unrelated change: while regenerating the doc, I noticed that Client.from_config_file was not included in the doc. I added a docstring to include it as well. This made me realize that the backend arg of the method made no sense since any config loaded from disk is necessarily a remote conf (otherwhise you can just do `Client(backend='local')`.


substra/sdk/client.py
```diff
@@ -116,19 +106,37 @@ def from_config_file(
         cls,
         profile_name: typing.Optional[str] = None,
```
Esadruhn, line 107: profile is required: `cfg.ConfigManager(config_path).get_profile(None)` raises ProfileNotFoundError
or use the DEFAULT_PROFILE_NAME variable in the _ProfileManager

substra/sdk/client.py
```diff
@@ -116,19 +106,37 @@ def from_config_file(
...
         config_path: typing.Union[str, pathlib.Path, None] = None,
-        user_path: typing.Union[str, pathlib.Path, None] = None,
+        tokens_path: typing.Union[str, pathlib.Path, None] = None,
         token: typing.Optional[str] = None,
         retry_timeout: int = DEFAULT_RETRY_TIMEOUT,
-        backend: str = 'remote',
     ):
-        cfg_manager = cfg.Manager(config_path)
-        profile = cfg_manager.from_config_file(name=profile_name)
+        """Returns a new Client configured with profile data from configuration files.
```
Esadruhn, line 113: Maybe mention that it is only for a remote backend ?

substra/sdk/config.py
```diff
@@ -12,27 +12,19 @@
...
 import logging
 import os
 
 logger = logging.getLogger(__name__)
 
 DEFAULT_PATH = os.path.expanduser('~/.substra')
+DEFAULT_TOKENS_PATH = os.path.expanduser('~/.substra-tokens')
 DEFAULT_VERSION = '0.0'
-DEFAULT_URL = 'http://127.0.0.1:8000'
 DEFAULT_PROFILE_NAME = 'default'
```
Esadruhn, line 25: see previous comment about profile_name being required, is this still useful ?

substra/sdk/client.py
jmorel: We could have the DEFAULT_PROFILE_NAME as default here. This way `Client.from_config_file()` returns the default profile of the default config files instead of raising an error.

substra/sdk/client.py
jmorel: Absolutely!

substra/sdk/config.py
jmorel: I think it is, otherwise when using the CLI it'll always be required to specify a profile name, which could be very annoying.

PR #202, opened 2020-07-15
Local backend - add and update compute plan

Esadruhn:
Add the add_compute_plan, update_compute_plan functions to the local backend.

## Main changes

- [x] add_compute_plan
- [x] update_compute_plan
- [ ] Hybrid debugging: run these on fake data (wait for https://github.com/SubstraFoundation/substra/pull/195) --> will be done afterwards in the hybrid debugging PR
- [x] add filter support on the `list` function (substra-tests uses it)
- [x] fix a bug on `AggregatetupleSpec`: the response object had an attribute `computePlanId` instead of `computePlanID`
- [x] Reordered the functions in backend.py by privacy (public vs private vs called only by private functions)

## Progress

`add_compute_plan` and `update_compute_plan` implemented, tests pass in substra-tests
Lack support for `clean_models`attribute --> raise an error if clean_models is True
Lack support for running on fake data --> will be implemented in the 'hybrid debugging' PR

## N.B.

Once the aggregate and composite tuples PR is merged, change the order of the functions to keep the public functions below.


Esadruhn:
Question: what does clean_models do ?

substra/sdk/backends/local/backend.py
```diff
@@ -739,8 +736,309 @@ def describe(self, asset_type, key):
...
+            except exceptions.NotFound:
+                pass
+        if traintuple_type == schemas.Type.Traintuple:
+            algo_type = schemas.Type.Algo
+        elif traintuple_type == schemas.Type.Aggregatetuple:
+            algo_type = schemas.Type.AggregateAlgo
+        elif traintuple_type == schemas.Type.CompositeTraintuple:
+            algo_type = schemas.Type.CompositeAlgo
+
+        algo = self._db.get(algo_type, traintuple.algo_key)
```
jmorel, line 758: This could probably be simplified a bit with the introduction of the algo_type attributes on the tuples.

substra/sdk/backends/local/backend.py
```diff
@@ -739,8 +736,309 @@ def describe(self, asset_type, key):
...
+        certified_testtuples.sort(key=lambda x: x['perf'], reverse=(sort == 'desc'))
+        board = {
+            'objective': objective.to_response(),
+            'testtuples': certified_testtuples
+        }
+        return board
 
     def cancel_compute_plan(self, compute_plan_id):
+        # Execution is synchronous in the local backend so this
+        # function does not make sense.
```
jmorel, line 789: 👍

substra/sdk/backends/local/backend.py
```diff
@@ -739,8 +736,309 @@ def describe(self, asset_type, key):
...
+                id_, rank = self.__get_id_rank_in_compute_plan(
+                    schemas.Type.Traintuple,
+                    key,
+                    compute_plan.id_to_key
+                )
+                visited[id_] = rank
+
+        while len(visited) != len(all_tuples):
+            node = set(all_tuples.keys()).difference(set(visited.keys())).pop()
+            self.__get_rank(node, visited, set(), all_tuples)
```
jmorel, line 838: I like this very much, it could also serve as the basis for the future autobatching feature for compute plan creation on remote backends. Do you think it could be moved to an independent function with dedicated tests?

substra/sdk/backends/local/backend.py
```diff
@@ -739,8 +736,309 @@ def describe(self, asset_type, key):
...
+            aggregatetuples,
+            compositetuples,
+            exist_ok,
+            spec_options
+            ):
+        for id_, rank in {
+            id_: rank
+            for id_, rank
+            in sorted(visited.items(), key=lambda item: item[1])
+        }.items():
```
jmorel, line 955: Isn't this equivalent to `for id_, rank in sorted(visited.items(), key=lambda item: item[1]):` ?

substra/sdk/backends/local/backend.py
```diff
@@ -739,8 +736,309 @@ def describe(self, asset_type, key):
...
+                    in_trunk_model_key=(compute_plan.id_to_key[compositetuple.in_trunk_model_id]
+                                        if compositetuple.in_trunk_model_id else None),
+                    out_trunk_model_permissions={
+                        "authorized_ids": compositetuple.out_trunk_model_permissions.authorized_ids
+                    },
+                    tag=compositetuple.tag,
+                    compute_plan_id=compute_plan.compute_plan_id,
+                    rank=rank,
+                    metadata=compositetuple.metadata
+                )
```
jmorel, line 1007: This is just an idea to make the __execute_compute_plan method smaller, not a requirement at all. 

We could have a class method schemas.CompositeTraintupleSpec.from_compute_plan_spec(compute_plan, compute_plan_tuple_spec) that would do all of this (same for the other tuples).

substra/sdk/backends/local/backend.py
Esadruhn: 😅

substra/sdk/backends/local/backend.py
Esadruhn: Indeed !

substra/sdk/backends/local/backend.py
Esadruhn: Yes, sure ! The function will contain all of this code:
```python
        while len(visited) != len(all_tuples):
            node = set(all_tuples.keys()).difference(set(visited.keys())).pop()
            self.__get_rank(node, visited, set(), all_tuples)
```
is that right ?

substra/sdk/backends/local/backend.py
Esadruhn: That's a good idea, the content of the loop looks like a big blob even though it is only creating the objects.

substra/sdk/backends/local/backend.py
jmorel: Yes, absolutely.

substra/sdk/backends/local/backend.py
```diff
@@ -739,8 +736,309 @@ def describe(self, asset_type, key):
...
+                testtuple_spec = schemas.TesttupleSpec(
+                    objective_key=testtuple.objective_key,
+                    traintuple_key=compute_plan.id_to_key[testtuple.traintuple_id],
+                    tag=testtuple.tag,
+                    data_manager_key=testtuple.data_manager_key,
+                    test_data_sample_keys=testtuple.test_data_sample_keys,
+                    metadata=testtuple.metadata,
+                )
+                testtuple = self.add(testtuple_spec, exist_ok, spec_options)
+                # TODO it seems that the testtuple are not in id_to_key ?
```
jmorel, line 1022: They are not :)

substra/sdk/backends/local/backend.py
Esadruhn: I put it in assets.py (and realised that the rest of the file was useless (did a search on all repos)), I do not know if it is the best place :/

Esadruhn: Thank you Travis, assets was used to generate cli.md😅

Esadruhn:
In the function substra/sdk/backends/local/backend.py - __create_compute_plan_from_tuple, do we trust the rank given by the user or do we recalculate it ? (case when an user creates a traintuple with compute_plan_id != None and gives a value for the rank)

Esadruhn:
I reordered the functions, tell me if it messes too much with the review, this is the latest commit so I can revert it easily

substra/sdk/schemas.py
```diff
@@ -130,6 +130,65 @@ def build_request_kwargs(self, local):
...
+class ComputePlanSpec(_BaseComputePlanSpec):
+    tag: Optional[str]
+    clean_models: Optional[bool]
+    metadata: Optional[Dict[str, str]]
+
+    type_: typing.ClassVar[Type] = Type.ComputePlan
+
+
+class UpdateComputePlanSpec(_BaseComputePlanSpec):
+    pass
```
jmorel, line 189: Aren't you missing the `type_` here?

tests/sdk/test_assets.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_compute_ranks(node_graph):
+    visited = assets.compute_ranks(node_graph=node_graph)
+    for key, rank in visited.items():
+        assert key == rank
+
+
+def test_compute_ranks_failure(node_graph):
+    node_graph[5].append(9)
+    with pytest.raises(exceptions.InvalidRequest):
```
jmorel, line 22: Could we have a more specific error here? Something dedicated to circular dependencies maybe.

substra/sdk/backends/local/backend.py
```diff
@@ -125,14 +125,9 @@ def __create_compute_plan_from_tuple(self, spec, key, in_tuples):
...
-                rank = 1 + max(
-                    [
-                        in_tuple.rank
-                        for in_tuple in in_tuples
-                        if in_tuple.compute_plan_id == compute_plan_id
-                    ]
-                )
+            else:
+                # Use the rank given by the user
+                rank = spec.rank
```
jmorel, line 130: We need to have a way to compute the default rank in case none was specified.

substra/sdk/schemas.py
Esadruhn: in the Type class, there is no type 'updateComputePlan', I thought that it was only for the types of stored objects

tests/sdk/test_assets.py
Esadruhn: If we want the tests on substra-tests to pass, the SDK must return an InvalidRequest error but we can also add a check on the message content

Extract from the test in substra-tests (substra-tests/tests/test_execution_compute_plan.py::test_compute_plan_circular_dependency_failure):
```
with pytest.raises(substra.exceptions.InvalidRequest) as e:
        client.add_compute_plan(cp_spec)

    assert 'missing dependency among inModels IDs' in str(e.value)
```

substra/sdk/backends/local/backend.py
Esadruhn: Good catch !

tests/sdk/test_assets.py
jmorel: Ok, let's leave it like this then, thanks.

substra/sdk/schemas.py
jmorel: Ok, having a type on this object is not strictly necessary, let's not do it then :)

PR #204, opened 2020-07-20
Update the compatibility table before release

Esadruhn:
Update the compatibility table: substra-tools 0.6.0 and substra 0.7.0-alpha.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,20 @@ CLI and SDK for interacting with Substra platform.
 
 ## Table of contents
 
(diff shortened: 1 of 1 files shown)


jmorel:
Since these are alpha releases I'm not sure we want to update the compatibility table. Alpha for me means not ready for prime time, only for people on the bleeding edge willing to deal with potential bugs and quirks. Putting it in the compatibility table is too much advertising for what is essentially a big work in progress.

Issue #205, opened 2020-07-21
[Feature request] Hybrid debugging


Esadruhn:
## Description

Create a "debugging" mode when using the Substra SDK. The principle is that the user only needs to change one parameter in his script to switch between the "normal" and "debugging" mode.

In this mode:
- adding an asset adds it locally
- fetching an asset fetches it either from the deployed platform or locally (local assets are identified by a 'key' that starts with `local_` or a similar prefix or suffix)
- executing a traintuple on a dataset fetched remotely has the following behaviour:
  - it counts the number of `train_samples_keys` of the dataset
  - it gets the same number of samples from the `fake` method of the opener
  - it executes on the fake data
- the `list` function lists the remote and local assets
- the `leaderboard` function gets the leaderboard from the deployed platform if the objective is on it, otherwise from the local data

### Limitation

A locally created traintuple, testtuple, aggregate tuple or composite traintuple cannot depend on a remote tuple.
Same for a compute plan: it is impossible to add a tuple to a compute plan that is on the deployed platform.

## Improvements

It must be clear to the user that executing a local tuple on a remote asset means that it uses the fake data. 
On the contrary, if the tuples executes on a local asset (the user created a dataset and data samples locally) then the real data is used.

--> the documentation should be clear about this
--> add a warning / print in the stdout when it uses the fake data ?

## Code example
```python
dataset_key = EXISTING_KEY
remote_traintuple_key = EXISTING_TRAINTUPLE_KEY

c = Client(<url>, <version>, debugging=True)
dataset = c.get_dataset(dataset_key)
data_samples = dataset['trainDataSampleKeys']

# Can also get the objective and algo from the deployed platform
c.add_objective(<objective>)
c.add_algo(<algo>)

c.add_traintuple({
        'train_data_sample_keys': [d['key'] for d in data_samples]
    })  # the traintuple executes on len(data_samples) fake data

# This fails:
c.add_traintuple({
        'in_models_keys': [remote_traintuple_key]
    }) 
```

natct10:
This would be awesome!

Esadruhn:
Closed with PR #206

PR #206, opened 2020-07-22
Debugging mode

Esadruhn:
#205 

## Progress

- [X] Remote dataset
- [X] Remote objective
- [X] Remote algo, aggregate algo, composite algo
- [X] Documentation and examples
- [X] Update `describe` functions
- [X] Update `download` functions
- [X] Update the `leaderboard` function

--> need some refacto and cleaning

### Remote vs local asset

- The local backend now has a 'remote backend' that it calls to get remote assets
- This is done by adding a new layer between the Backend and the Local DB: the DataAccess layer, that decides whether to call the Local DB or the remote backend (the name of the layer may be changed)
- [x] How to know whether an asset is local or remote: test if its key starts with 'local_'. 
We could test on the length of the key but it is not the same for the compute plan and the other assets:
  - if the asset is not a compute plan, 64 for remote, 70 for local
  - if the asset is a compute plan, ??? for remote, 38 for local (need to check for remote)

### Models : use the same object no matter the origin of the asset

- Use "models" to define the model of remote and local assets:
  - [X] Convert the response from remote to snake_case
  - [X] Update the models since there are slight differences between the local and remote objects
  - [x] Convert the dictionary to a 'model': we need a mapping between schemas.Type and Model, for now it is the `SCHEMA_TO_MODEL` dictionary in models.py --> **is it the best way ?**
- [x] in the worker: download the asset from the remote for the execution. This does not put the asset into the local db. If the files already exists, it uses them (is it OK ? Is there a risk they are updated ?)
- [x] Downloading the assets: had to change the function to uncompress archive: we cannot know the compression method from the name of the archive ('.zip' or '.tar.gz') since we renamed it when downloading it

### Fake data

- [x] in substra tools, n_samples should be mandatory for the 'fake' functions: https://github.com/SubstraFoundation/substra-tools/pull/42
- [x] substra_tools and substra: for the compute metrics command, do we need n_fake_samples ?

## Substra tests

https://github.com/SubstraFoundation/substra-tests/pull/123


Esadruhn:
- Works with titanic when changing client, debug=True in the add_train_algo_constant.py script
- DAL needs some refacto (we can certainly have only one 'download' function) and getting / downloading an `algo` from the deployed platform does not work
- Rename dal.py and DataAccess class ?

substra/sdk/backends/local/models.py
```diff
@@ -109,17 +127,22 @@ def __str__(self):
 
 
 class DataSample(_Model):
```
Esadruhn, line 129: Are the changes logic ?
- we could instead remove "owner"
- do the data sample keep the list of the data manager keys ? The remote backend does not return it

substra/sdk/client.py
```diff
@@ -52,40 +52,70 @@ def wrapper(*args, **kwargs):
...
-            self._backend_name,
-            url=self._url,
-            version=self._version,
-            insecure=self._insecure,
-            token=self._token,
-            retry_timeout=self._retry_timeout,
-        )
+        self._backend = self._get_backend(debug)
+
+    def _get_backend(self, debug: bool):
```
Esadruhn, line 102: add comment

Esadruhn: Updated !

references/sdk.md
```diff
@@ -2,12 +2,35 @@
...
-Set the `backend` to 'local' to use the local debugging.
-One client corresponds to one profile.
+Args:
+
+    url (typing.Optional[str], optional): URL of the Substra platform. Mandatory
+        to connect to a Substra platform. If no URL is given and debug is True, all
+        assets must be created locally.
+        Defaults to None.
+
+    token (typing.Optional[str], optional): Token to authenticate to the Substra platform.
```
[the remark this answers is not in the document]
Esadruhn: Yes, the doctsring is automatically generated and copies the typing. I changed it, indeed for the user `str, optional` is clear enough I think

substra/sdk/client.py
```diff
@@ -52,40 +52,70 @@ def wrapper(*args, **kwargs):
...
+
+        debug (bool, optional): Whether to use the default or debug mode.
+            In debug mode, new assets are created locally but can access assets from
+            the deployed Substra platform. The platform is in read-only mode.
+            Defaults to False.
     """
 
     def __init__(
         self,
         url: typing.Optional[str] = None,
```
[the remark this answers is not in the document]
Esadruhn: I think it is a shame to let go of the fully local backend when it is so simple to keep it as is, but I am not sure how useful it will be (and I am definitely partial on this !), maybe ask the users ?

substra/sdk/client.py
jmorel: Yes, thank you!

docs/debugging.md
```diff
@@ -0,0 +1,117 @@
...
+      - [Access the logs](#access-the-logs)
+      - [Debug with pdb](#debug-with-pdb)
+
+## Dependencies
+
+The required dependencies are:
+
+- Docker:
+  - Mac: [Docker Desktop](https://www.docker.com/products/docker-desktop)
+  - Ubuntu/Debian: `sudo apt install docker docker-compose`
```
samlesu, line 20: docker-compose could be removed?

docs/debugging.md
```diff
@@ -0,0 +1,117 @@
...
+- Python 3 (recommended 3.6 or 3.7)
+  - It is recommended to use a virtual environment to install Substra packages (for instance [virtualenv](https://virtualenv.pypa.io/en/latest/))
+- Substra package
```
samlesu, line 23: This is already a requirement for running substra so I would keep only stuff specific to the debugging mode.

docs/debugging.md
```diff
@@ -0,0 +1,117 @@
...
+    - [Dependency between assets](#dependency-between-assets)
+    - [Debugging with logs and breakpoints](#debugging-with-logs-and-breakpoints)
+      - [Access the logs](#access-the-logs)
+      - [Debug with pdb](#debug-with-pdb)
+
+## Dependencies
+
+The required dependencies are:
+
+- Docker:
```
samlesu, line 18: Maybe explain why this is needed?

docs/debugging.md
```diff
@@ -0,0 +1,117 @@
...
+### Use assets from a deployed Substra platform and local assets
+
+With this, you gain access to the platform in 'read-only' mode and any asset you create is created locally.
+
+This means that any function to get, describe or download an asset works with assets from the platform and local assets. Functions to
+list assets will list the assets from the platform and the local ones.  
+Functions that create a new asset (they start with `add`) only create local assets.  
+To differentiate between a local asset and a remote one, look at their key (or pkhash): for local assets, it starts with `local_`.
+
+Local assets are saved in-memory, so they are deleted at the end of the script.
```
samlesu, line 46: We could add that the local assets have the same lifetime as the related Client?

docs/debugging.md
```diff
@@ -0,0 +1,117 @@
...
+
+Any asset created locally can depend on an asset from the platform or on another local asset:
+- a traintuple can use an algo and dataset from the platform
+- a testtuple can use an objective from the platform
+- etc.
+
+There is one exception: a traintuple, testtuple or any other tuple created locally cannot depend on a tuple from the platform 
+or from a compute plan defined on the platform. This is impossible because it is forbidden to download models from the platform.
+
+Since no data can leave the platform, a task depending on a dataset from the platform uses the fake data generated from the opener. The
```
samlesu, line 58: 👍

docs/debugging.md
```diff
@@ -0,0 +1,117 @@
...
+print(traintuple["log"])
+```
+
+Moreover, if there is an error in the execution, those logs are displayed in the call stack.  
+So any `print` or warning in the algo file for example will be displayed.
+
+#### Debug with pdb
+
+The execution is done synchronously, so the script waits for the task in progress to end before continuing.  
+The execution of the tuples happens in Docker containers that are spawned on the fly and removed once the execution is done.
```
samlesu, line 81: I think these 2 lines are really important and should be in a more common section (in "Run your code locally" maybe?).

examples/debugging/scripts/hybrid_debugging.py
```diff
@@ -0,0 +1,114 @@
+# Copyright 2018 Owkin, inc.
```
samlesu, line 1: could be renamed without the word "hybrid"?

references/sdk.md
```diff
@@ -2,12 +2,35 @@
...
+        Defaults to None.
+
+    retry_timeout (int, optional): Number of seconds before attempting a retry call in case
+        of timeout.
+        Defaults to 5 minutes.
+
+    version (str, optional): API version, do not change it. Defaults to '0.0'.
+
+    insecure (bool, optional): If True, the client can call a not-certifed backend. This is
+        for debugging purposes.
```
samlesu, line 27: development purposes? I would avoid the word debugging to not confuse with the next `debug` option.

substra/sdk/backends/local/backend.py
```diff
@@ -252,22 +262,13 @@ def __execute_compute_plan(
         return compute_plan
 
     def __format_for_leaderboard(self, testtuple):
```
samlesu, line 264: This is a bit too late, but IMHO we don't need to support the leaderboard function in the debugging mode. As it is already done, we can keep it like this :)

substra/sdk/backends/local/backend.py
```diff
@@ -328,9 +339,10 @@ def _add_composite_algo(self, spec, exist_ok, spec_options=None):
     def _add_dataset(self, spec, exist_ok, spec_options):
         self.__check_metadata(spec.metadata)
         permissions = self.__compute_permissions(spec.permissions)
+        key = fs.hash_file(spec.data_opener)
         asset = models.Dataset(
-            key=fs.hash_file(spec.data_opener),
-            pkhash=fs.hash_file(spec.data_opener),
+            key=key,
+            pkhash=key,
```
samlesu, line 345: Why do we need `key` and `pkhash` fields?

substra/sdk/backends/local/backend.py
```diff
@@ -95,7 +105,7 @@ def __add_compute_plan(
             ]
         ])
         compute_plan = models.ComputePlan(
-            compute_plan_id=uuid.uuid4().hex,
+            compute_plan_id="local_" + uuid.uuid4().hex,
```
samlesu, line 108: Could we define "local_" in a global constant? 

This would allow to easily change the constant value without forgetting an occurence.

substra/sdk/backends/local/compute/spawner.py
```diff
@@ -36,14 +36,16 @@ def _unzip(archive, to_):
...
+    success = False
+    errors = list()
+    for handler in [_unzip, _untar]:
+        try:
+            handler(archive, to_)
+            success = True
+        except Exception as e:
+            errors.append(e)
+    if not success:
+        raise exceptions.InvalidRequest(f"Cannot uncompress '{archive}', error: \n{errors}", 400)
```
samlesu, line 48: What's the purpose of modifying this code?

The main drawback of the new code is that we will always run both handlers.

One enhancement could be to catch 2 different kind of errors:
- "Invalid archive extensions"
- "Cannot uncompress"
?

substra/sdk/backends/local/compute/worker.py
```diff
@@ -55,19 +55,26 @@ def _get_address_in_container(model_key, volume, container_volume):
...
-        self._db = db.get()
+        self._db = db
         self._spawner = spawner.get()
 
+    @staticmethod
+    def _is_local(key: str):
+        """Check if the key corresponds to a local
+        or remote asset.
+        """
+        return key.startswith("local_")
```
samlesu, line 68: We should use here a constant for "local_".
As this method is duplicated, maybe we could move this method to the DataAccess key? And move inside this class a helper to generate a local key?

substra/sdk/backends/local/dal.py
```diff
@@ -0,0 +1,137 @@
...
+class DataAccess:
+    """Data access layer.
+
+    This is an intermediate layer between the backend and the local/remote data access.
+    """
+
+    def __init__(self, remote_backend: typing.Optional[backend.Remote]):
+        self._db = db.get()
+        self._remote = remote_backend
+        self._tmp_dir = tempfile.TemporaryDirectory(prefix="/tmp/")
```
samlesu, line 32: With the new feature to save the local files in a working directory, we will end up with 2 tmpdirs. It would be good to have a unique tempdir.

Maybe we could move the __copy_file function to this class?
It would make sense, this classs would be responsible for storing both the assets and their content.

substra/sdk/backends/local/dal.py
```diff
@@ -0,0 +1,137 @@
...
+        return pathlib.Path(self._tmp_dir.name)
+
+    @staticmethod
+    def _is_local(key: str):
+        """Check if the key corresponds to a local
+        or remote asset.
+        """
+        return key.startswith("local_")
+
+    def _get_asset_name(self, type_):
```
samlesu, line 45: Maybe get_asset_content_filename would be more descriptive?

substra/sdk/backends/local/dal.py
```diff
@@ -0,0 +1,137 @@
...
+                pathlib.Path.mkdir(tmp_directory)
+
+                self._remote.download(
+                    type_,
+                    field_name + ".storageAddress",
+                    key,
+                    asset_path,
+                )
+
+                description = self._remote.describe(type_, key)
```
samlesu, line 107: We should ignore the asset description as it's not needed for its execution.

substra/sdk/backends/local/hasher.py
```diff
@@ -14,4 +14,4 @@ def update(self, v):
         self._h.update(v)
 
     def compute(self):
-        return self._h.hexdigest()
+        return "local_" + self._h.hexdigest()
```
samlesu, line 17: IMO the "local_" keyword should be added outside of the hasher module. The hasher module should be only responsible for hashing stuff.

substra/sdk/backends/local/models.py
```diff
@@ -16,13 +16,26 @@
...
+UriPath = Union[FilePath, AnyUrl, str]
+
+CAMEL_TO_SNAKE_PATTERN = re.compile(r'(.)([A-Z][a-z]+)')
+CAMEL_TO_SNAKE_PATTERN_2 = re.compile(r'([a-z0-9])([A-Z])')
+
+
+def to_snake_case(camel_str):
+    name = CAMEL_TO_SNAKE_PATTERN.sub(r'\1_\2', camel_str)
+    name = CAMEL_TO_SNAKE_PATTERN_2.sub(r'\1_\2', name).lower()
+    return name.replace('_i_d', '_id')
```
samlesu, line 37: 💯

substra/sdk/backends/local/models.py
samlesu: Yes, in this case we could remove it.

substra/sdk/client.py
```diff
@@ -52,40 +52,77 @@ def wrapper(*args, **kwargs):
...
+        self._backend = self._get_backend(debug)
+
+    def _get_backend(self, debug: bool):
+        # Three possibilites:
+        # - debug is False: get a remote backend
+        # - debug is True and no url is defined: fully local backend
+        # - debug is True and url is defined: local backend that connects to
+        #                           a remote backend (read-only)
+        backend = None
+        if (debug and self._url) or not debug:
```
samlesu, line 109: Could it be simplified to `if self._url` ?

substra/sdk/backends/local/backend.py
```diff
@@ -79,6 +82,13 @@ def __compute_permissions(permissions):
             permissions.authorized_ids.append(_BACKEND_ID)
         return permissions
 
+    @staticmethod
+    def _is_local(key: str):
+        """Check if la clé correspond à un asset local
+        ou d'une plateforme déployée.
+        """
```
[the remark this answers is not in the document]
Esadruhn: Good catch 😅

examples/debugging/scripts/hybrid_debugging.py
Esadruhn: Indeed, is `debugging.py` OK with you ? In the sense that we have a 'debug' mode and a specific use of this mode is the 'fully local' debugging.

substra/sdk/backends/local/compute/spawner.py
Esadruhn: Yes, if we could trust the extension (.zip or .tar.gz) of the archive to know which handler to use it would be better. The problem is that when we download the archive from the remote, we rename it and lose the extension, so we have to try both handlers.

Catching their specific errors is a good idea !

substra/sdk/backends/local/compute/worker.py
Esadruhn: You're right, DataAccess should contain everything related to local vs remote

The DataAccess class now has 2 functions:
- is_local --> check if the key starts with 'local_'
- get_local_key --> add 'local_' in front of the key

What do you think ?

substra/sdk/backends/local/dal.py
Esadruhn: Yes, better !

substra/sdk/backends/local/dal.py
Esadruhn: One less call to the remote backend and a faster execution :)

substra/sdk/backends/local/hasher.py
Esadruhn: I moved everything to DataAccess

substra/sdk/client.py
Esadruhn: It would work, I made this condition to keep the same behaviour as before in the 'normal' (ie not debug) case.  

If the condition is `if self._url:`, then the client is created without error but any subsequent call fails with a NoneType error (since client._backend is None)

With the condition `if (debug and self._url) or not debug:`, in normal mode if there is no URL the client creation fails with an explicit error `substra.sdk.exceptions.SDKException: url required to connect to the Substra server`

docs/debugging.md
Esadruhn: You are right, I only left that Substra must be installed and the Docker requirement.

docs/debugging.md
Esadruhn: OK !

substra/sdk/backends/local/compute/spawner.py
Esadruhn: After a quick verification, the tarfile and zipfile packages provide `is_tarfile` and `is_zipfile` functions, which is better than the `try...except...` that I used before !

substra/sdk/backends/local/backend.py
Esadruhn: I thought that the remote Dataset returned key and pkhash, actually it has only key, changed it :)

Esadruhn: But the testDataSampleKeys and trainDataSampleKeys can be None, so in models.py I put a default value of `list()` for these fields instead of making them optional.

Esadruhn: Thanks, because of your comment I realised that many of the `list_` functions failed, the behaviour with the `pkhash` not defined is actually common to all objects that have a pkhash.

substra/sdk/backends/local/dal.py
Esadruhn: Good idea, I answered on the "copy assets" PR

examples/debugging/scripts/hybrid_debugging.py
samlesu: Perfect, thanks!

substra/sdk/backends/local/compute/worker.py
samlesu: Sounds good, thanks.

substra/sdk/backends/local/compute/worker.py
```diff
@@ -55,9 +55,9 @@ def _get_address_in_container(model_key, volume, container_volume):
 class Worker:
     """ML Worker."""
 
-    def __init__(self):
+    def __init__(self, db: dal.DataAccess):
```
samlesu, line 58: 💯

substra/sdk/client.py
samlesu: Thanks for the explanation. Looks good :)

substra/sdk/backends/local/compute/spawner.py
samlesu: Looks great!

substra/sdk/backends/local/backend.py
samlesu: If possible, it would like to avoid having a single `pkhash` field in the SDK responses. This field is in my opinion internal to the backend and should not be exposed to the user. One field for the asset unique id is enough :)

But as the backend have inconsistent responses, I'm not sure if it's doable :)

Esadruhn: Actually, where does it come from to have both `pkhash` and `key` ? I thought that `pkhash` was in the process of being renamed `key` but it does not seem to be the case.
I can always hide it from the users (`to_response` method), but they have it in 'normal' mode and I know that at least Jean's code relies on it rather than `key`

samlesu: > Actually, where does it come from to have both pkhash and key ?

I think it comes from having responses returned by both the backend and the chaincode and multiple refactoring. Pkhash is internal to django (or DRF), it's an abbreviation for "primary key hash".
If I remember well, `pkhash` was used in the old times and a switch to `key` has been made progressively.

Let's keep it that way, later we could a proper refactoring in the backend / sdk to ensure pkhash is not visible from the user point of view.

substra/sdk/client.py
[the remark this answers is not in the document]
samlesu: Having an optional url seems good to me. 
Maybe we could make the docstring clearer:
 
> If no URL is given and debug is True, all assets must be created locally.

to 

> If no URL is given and debug must be defined as True, all assets will be created locally.

PR #215, opened 2020-08-13
Remove the 'version' from the client

### LINKED ISSUE #208: [Feature request] Remove the 'version' from the client
## Current

There is a parameter `version` in the `__init__` and `from_config_file` functions of the Client. This parameters refers to a versioning to the API, but there is only one version available (and it is not in the works to change that)

## What the feature changes

Remove 'version' from those function's signatures, hardcode it where it is needed.  
It removes a confusing parameter for the user.

maeldebon:
Fixes #208 

This PR aims to remove the 'version' parameter from the client since there is only one API version available.


setup.py
```diff
@@ -25,14 +25,9 @@
...
-about = {}
-with open(os.path.join(here, 'substra', '__version__.py'), 'r', 'utf-8') as fp:
-    exec(fp.read(), about)
-
-
 setup(
     name='substra',
-    version=about['__version__'],
+    version='0.0',
     description='Substra CLI for interacting with substra-backend',
```
jmorel, line 31: This is not the same version, this refers to the release of the substra lib, not to the version of the backend API. We should keep the __version__.py file and its uses.

substra/sdk/client.py
```diff
@@ -110,7 +106,7 @@ def _get_backend(self, debug: bool):
             backend = backends.get(
                 "remote",
                 url=self._url,
-                version=self._version,
+                version='0.0',
```
jmorel, line 109: I think you should remove the version arg of the backends.get function, and hardcode the 0.0 in the rest_client directly.

substra/__init__.py
```diff
@@ -12,12 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from substra.__version__ import __version__
```
samlesu, line 15: This is unrelated to the API version. This is the version of the package and should not be removed.

substra/cli/interface.py
```diff
@@ -263,7 +263,6 @@ def wrapper(*args, **kwargs):
 
 
 @click.group()
-@click.version_option(__version__)
```
samlesu, line 266: Same comment here.

tests/test_cli.py
```diff
@@ -64,11 +64,6 @@ def test_command_help():
     assert 'Usage:' in output
 
 
-def test_command_version():
```
samlesu, line 67: Same here.

substra/__init__.py
maeldebon: Oops, I forgot to put it back when I fixed jmorel's comments.

substra/sdk/backends/remote/rest_client.py
```diff
@@ -24,13 +24,13 @@
...
 
-    def __init__(self, url, version, insecure, token):
+    def __init__(self, url, insecure, token):
         self._default_kwargs = {
             'verify': not insecure,
         }
         self._headers = {
             'Authorization': f"Token {token}",
-            'Accept': f'application/json;version={version}',
+            'Accept': 'application/json',
```
jmorel, line 33: Did you try running substra-tests with this code? I think it should fail because the version is missing. You should instead write `'Accept': 'application/json;version=0.0',`

substra/__init__.py
```diff
@@ -12,8 +12,8 @@
...
 from substra.sdk import Client, exceptions
+from substra.__version__ import __version__
```
jmorel, line 16: Minor: whenever you change something and then undo the change, you should try to undo the change completely and not just undo the functionality change. Here for example we have 2 additional files in the diff that shouldn't be here at all (setup.py and substra/__init__.py) because the changes you made neither change the functionnality or fix/improve formatting. One handy solution to revert all changes made to a file is to check it out from the master branch:

```sh
git checkout master -- substra/__init__.py
git checkout master -- setup.py
```

maeldebon: Yes, I should have done that, I'll do it!

substra/sdk/backends/remote/rest_client.py
maeldebon: I'll fix this.

maeldebon:
> LGTM! By the way, I see you wrote "Related to #208" in the description. If the goal was for the issue to be closed when the PR is merged, you should instead write "Fixes #208". See https://docs.github.com/en/enterprise/2.16/user/github/managing-your-work-on-github/closing-issues-using-keywords

Yes indeed, I'll do that now!

PR #216, opened 2020-08-13
Compute plan auto batching

Esadruhn:
Remote backend, when adding or updating a compute plan, make it by batch to avoid overloading the server (and getting timeouts)

## Changes

- Move the tuple parsing function to utils
- remote/backend.py: special case when adding a compute plan, first we create a compute plan with part of the tuples then we update it with consecutive calls
- remote/backend.py: the `update_compute_plan` function is done in several calls
- Batch size: 20
- option to deactivate auto batching and change the batch size

Companion PR: https://github.com/SubstraFoundation/substra-tests/pull/136

### DIFF
--- a/substra/sdk/backends/local/dal.py
+++ b/substra/sdk/backends/local/dal.py
@@ -109,7 +109,6 @@ class DataAccess:
             if not tmp_directory.exists():
                 pathlib.Path.mkdir(tmp_directory)
 
-            if not asset_path.exists():
                 self._remote.download(
                     type_,
                     field_name + ".storageAddress",
--- a/substra/sdk/backends/local/dal.py
+++ b/substra/sdk/backends/local/dal.py
@@ -109,6 +109,7 @@ class DataAccess:
             if not tmp_directory.exists():
                 pathlib.Path.mkdir(tmp_directory)
 
+            if not asset_path.exists():
                 self._remote.download(
                     type_,
                     field_name + ".storageAddress",
--- a/substra/sdk/backends/remote/backend.py
+++ b/substra/sdk/backends/remote/backend.py
@@ -141,6 +141,8 @@ class Remote(base.BaseBackend):
                                     spec_options=None):
         """Auto batching of the compute plan tuples
         """
+        spec_options = spec_options or dict()
+
         # Create the dependency graph and get the dict
         # of tuples by id
         (
@@ -177,6 +179,19 @@ class Remote(base.BaseBackend):
 
         # Sort the tuples by rank
         sorted_by_rank = sorted(visited.items(), key=lambda item: item[1])
+
+        # Special case: no tuples
+        if len(sorted_by_rank) == 0:
+            spec_options.setdefault(AUTO_BATCHING, False)
+            spec_options[AUTO_BATCHING] = False
+            if compute_plan_id:
+                return self.update_compute_plan(
+                    compute_plan_id=compute_plan_id,
+                    spec=spec,
+                    spec_options=spec_options
+                )
+            else:
+                return self.add(spec=spec, exist_ok=exist_ok, spec_options=spec_options)
 
         # Create / update by batch
         for i in range(math.ceil(len(sorted_by_rank) / COMPUTE_PLAN_BATCH_SIZE)):
--- a/substra/sdk/backends/remote/backend.py
+++ b/substra/sdk/backends/remote/backend.py
@@ -153,20 +153,18 @@ class Remote(base.BaseBackend):
         ) = spec.get_dependency_graph()
 
         compute_plan = None
-        visited = dict()
-        already_created_keys = dict()
+        already_created_keys = set()
         if compute_plan_id:
             # Here we get the pre-existing tuples and assign them the minimal rank
             for dependencies in tuple_graph.values():
                 for dependency_key in dependencies:
                     if dependency_key not in tuple_graph:
-                        already_created_keys[dependency_key] = 0
+                        already_created_keys.add(dependency_key)
 
         # Compute the relative ranks of the new tuples (relatively to each other, these
         # are not their actual ranks in the compute plan)
         visited = graph.compute_ranks(
             node_graph=tuple_graph,
-            visited=visited,
             node_to_ignore=already_created_keys
         )
 
--- a/substra/sdk/backends/remote/backend.py
+++ b/substra/sdk/backends/remote/backend.py
@@ -14,19 +14,17 @@
 
 from copy import deepcopy
 import logging
-import math
 import json
-import os
 
-from substra.sdk import exceptions, schemas, graph
+from substra.sdk import exceptions, schemas, compute_plan
 from substra.sdk.backends import base
 from substra.sdk.backends.remote import rest_client
 
 logger = logging.getLogger(__name__)
 
 DEFAULT_RETRY_TIMEOUT = 5 * 60
-DEFAULT_COMPUTE_PLAN_BATCH_SIZE = 10000000000
 AUTO_BATCHING = "auto_batching"
+BATCH_SIZE = 'batch_size'
 
 
 def _get_asset_key(data):
@@ -105,6 +103,8 @@ class Remote(base.BaseBackend):
         """Add an asset."""
         spec_options = spec_options or {}
         asset_type = spec.__class__.type_
+        # Remove batch_size from spec_options
+        batch_size = spec_options.pop(BATCH_SIZE, None)
 
         if asset_type == schemas.Type.DataSample:
             # data sample corner case
@@ -113,13 +113,13 @@ class Remote(base.BaseBackend):
                 exist_ok=exist_ok,
                 spec_options=spec_options,
             )
-        elif asset_type == schemas.Type.ComputePlan and \
-                (AUTO_BATCHING not in spec_options or spec_options[AUTO_BATCHING]):
-            spec_options.pop(AUTO_BATCHING, None)
+        elif asset_type == schemas.Type.ComputePlan and spec_options.pop(AUTO_BATCHING, False):
+            assert batch_size
             # Compute plan auto batching feature
             return self._auto_batching_compute_plan(
                 spec=spec,
                 exist_ok=exist_ok,
+                batch_size=batch_size,
                 spec_options=spec_options,
             )
 
@@ -137,52 +137,23 @@ class Remote(base.BaseBackend):
 
     def _auto_batching_compute_plan(self,
                                     spec,
+                                    batch_size,
                                     compute_plan_id=None,
                                     exist_ok=None,
                                     spec_options=None):
         """Auto batching of the compute plan tuples
         """
         spec_options = spec_options or dict()
-
-        # Create the dependency graph and get the dict
-        # of tuples by id
-        (
-            tuple_graph,
-            traintuples,
-            aggregatetuples,
-            composite_traintuples
-        ) = spec.get_dependency_graph()
-
-        compute_plan = None
-        already_created_keys = set()
-        if compute_plan_id:
-            # Here we get the pre-existing tuples and assign them the minimal rank
-            for dependencies in tuple_graph.values():
-                for dependency_key in dependencies:
-                    if dependency_key not in tuple_graph:
-                        already_created_keys.add(dependency_key)
-
-        # Compute the relative ranks of the new tuples (relatively to each other, these
-        # are not their actual ranks in the compute plan)
-        visited = graph.compute_ranks(
-            node_graph=tuple_graph,
-            node_to_ignore=already_created_keys
+        batches = compute_plan.auto_batching(
+            spec,
+            is_creation=compute_plan_id is None,
+            batch_size=batch_size,
         )
 
-        # Add the testtuples to 'visited' to take them into account in the batches
-        testtuples = dict()
-        if spec.testtuples:
-            for testtuple in spec.testtuples:
-                visited['test_' + testtuple.traintuple_id] = visited[testtuple.traintuple_id] + 1
-                testtuples['test_' + testtuple.traintuple_id] = testtuple
-
-        # Sort the tuples by rank
-        sorted_by_rank = sorted(visited.items(), key=lambda item: item[1])
-
         # Special case: no tuples
-        if len(sorted_by_rank) == 0:
-            spec_options.setdefault(AUTO_BATCHING, False)
+        if batches is None:
             spec_options[AUTO_BATCHING] = False
+            spec_options[BATCH_SIZE] = batch_size
             if compute_plan_id:
                 return self.update_compute_plan(
                     compute_plan_id=compute_plan_id,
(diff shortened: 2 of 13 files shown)


substra/sdk/backends/remote/backend.py
```diff
@@ -109,24 +111,134 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+    ):
+        for elem_id, _ in tuple_graph:
+            if elem_id in traintuples:
+                spec.traintuples.append(traintuples[elem_id])
+            elif elem_id in aggregatetuples:
+                spec.aggregatetuples.append(aggregatetuples[elem_id])
+            elif elem_id in composite_traintuples:
+                spec.composite_traintuples.append(composite_traintuples[elem_id])
+            else:
+                raise Exception("ERROR")
```
samlesu, line 147: We should avoid raising errors using the `Exception`. It is a good practice to subclass the base class to create our own errors. Such exception classes already exist in the substra SDK and should be use. It allows the end user our the dev to catch it and to handle it gracefully if needed.

It seems that this exception should never occur; or if it is raised it will highlight a developper bug. If this is the case, using an assert would be appropriate.

substra/sdk/backends/remote/backend.py
```diff
@@ -109,24 +111,155 @@ def add(self, spec, exist_ok=False, spec_options=None):
                 exist_ok=exist_ok,
                 spec_options=spec_options,
             )
+        elif asset_type == schemas.Type.ComputePlan:
```
jmorel, line 114: I think there should be an option to deactivate the autobatching feature just in case :)

substra/sdk/schemas.py
```diff
@@ -183,6 +183,45 @@ class _BaseComputePlanSpec(_Spec, abc.ABC):
...
 
+    def get_dependency_graph(self):
+        """Get the tuple dependency graph and, for each type of tuple, a mapping table id/tuple.
+        """
+        tuple_graph = dict()
+        traintuples = dict()
+        if self.traintuples:
+            for traintuple in self.traintuples:
+                if traintuple.in_models_ids is not None:
+                    tuple_graph[traintuple.traintuple_id] = traintuple.in_models_ids
```
jmorel, line 194: What do you think of adding a check making sure that all IDs are unique? It could be checking that traintuple.traintuple_id is not already in tuple_graph for example. This could prevent some really weird behavior.

substra/sdk/backends/remote/backend.py
```diff
@@ -109,24 +111,155 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+            tuple_graph,
+            traintuples,
+            aggregatetuples,
+            composite_traintuples
+        ) = spec.get_dependency_graph()
+
+        compute_plan = None
+        visited = dict()
+        existing_keys = dict()
+        if compute_plan_id is not None:
```
jmorel, line 174: We could use `if compute_plan_id:` here, it would be a bit broader to trigger and easier to read.

substra/sdk/backends/remote/backend.py
```diff
@@ -109,24 +111,155 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+            # If we are updating an existing compute plan, we need the
+            # existing tuples to compute the new tuples ranks.
+            compute_plan = self.get(schemas.Type.ComputePlan, compute_plan_id)
+            # TODO find a good way of doing this
+            # We need the id and rank of each tuple that is in the dependency graph
+            # and already exists --> too long to get each existing tuple from the backend !!!
+            # Temporary solution: all new tuples
+            # have a rank equal (a minima) to nb_existing_tuples + 1
+            min_rank = compute_plan["tupleCount"]
+            existing_keys = {key: min_rank for dependencies in tuple_graph.values() for key in dependencies if key not in tuple_graph}
```
jmorel, line 184: I'm sorry, but two for loops with a conditional one is too hard to read for me. Could you break it down?

Also, it says "existing keys" and "key" but it seems to really be iterating over the ids of yet-to-created tuples. Could you update the naming to reflect this (if it is indeed the case) ?

substra/sdk/backends/remote/backend.py
Esadruhn: Indeed, this was for debugging purposes, I remove it.

substra/sdk/backends/remote/backend.py
```diff
@@ -109,24 +111,120 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+        # Auto batching of the compute plan tuples
+        # if compute_plan_id is None, we first create the compute plan
+        (
+            tuple_graph,
+            traintuples,
+            aggregatetuples,
+            composite_traintuples
+        ) = spec.get_dependency_graph()
+        visited = utils.compute_ranks(node_graph=tuple_graph)
+        sorted_by_rank = sorted(visited.items(), key=lambda item: item[1])
```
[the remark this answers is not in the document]
Esadruhn: I created a graph module that handles all these operations

substra/sdk/backends/remote/backend.py
```diff
@@ -109,24 +111,120 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+                tmp_spec.aggregatetuples = list()
+                tmp_spec.testtuples = list()
+                self.__fill_tuple_lists(
+                    tmp_spec,
+                    sorted_by_rank[start:end],
+                    traintuples,
+                    aggregatetuples,
+                    composite_traintuples,
+                    testtuples_by_train_id
+                )
```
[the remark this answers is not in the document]
Esadruhn: I moved the `fill_tuple_lists` function to the graph module and changed a bit how it assigned values, is it better ?

substra/sdk/schemas.py
Esadruhn: yes, checking that the user did not give two tuples with the same id makes sense ! I think it is checked server side too but I don't know what kind of behaviour it might induce in the auto batching 👍

substra/sdk/backends/remote/backend.py
Esadruhn: Broke it down, it was not very readable ! 
It is the keys of the pre-existing tuples that we are getting, the "tuple_graph" is built with the keys of the new tuples. I also changed how they were managed to avoid adding them to 2 dicts then removing them, tell me if it is better.

substra/sdk/backends/remote/backend.py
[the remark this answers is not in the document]
samlesu: I think so yes, thanks!

Esadruhn:
Changed the code to count the testtuples in the batches (tests in progress)

substra/sdk/backends/remote/backend.py
Esadruhn: Added an option ! The most efficient way to batch will always be if the user does it since he does not have to compute the dependency graph.

substra/sdk/schemas.py
```diff
@@ -183,6 +183,51 @@ class _BaseComputePlanSpec(_Spec, abc.ABC):
     aggregatetuples: Optional[List[ComputePlanAggregatetupleSpec]]
     testtuples: Optional[List[ComputePlanTesttupleSpec]]
 
+    def get_dependency_graph(self):
```
jmorel, line 186: What do you think of creating a `_insert_into_graph(tuple_graph, tuple_id, in_model_ids)` function that could then be used by all the if clauses? I think it'd remove a bit of code that's almost duplicated but not really.

Esadruhn:
Re-arranged following discussion with @jmorel 
Still have to test a few edge cases (e.g. an update with a testtuple that depends on an existing tuple) and the CI fails, but almost everything is here

substra/sdk/backends/remote/backend.py
```diff
@@ -109,26 +113,99 @@ def add(self, spec, exist_ok=False, spec_options=None):
                 exist_ok=exist_ok,
                 spec_options=spec_options,
             )
+        elif asset_type == schemas.Type.ComputePlan and spec_options.pop(AUTO_BATCHING, False):
+            assert batch_size
```
jmorel, line 117: Could you replace this with an exception?

substra/sdk/graph.py
```diff
@@ -0,0 +1,131 @@
...
+                raise exceptions.InvalidRequest(
+                    "missing dependency among inModels IDs", 400
+                )
+            else:
+                edges.add(edge)
+
+    return ranks
+
+
+def filter_tuples_in_list(
```
jmorel, line 103: I think this method should be in compute_plan.py

substra/sdk/compute_plan.py
```diff
@@ -0,0 +1,157 @@
...
+        composite_traintuples,
+    ) = get_dependency_graph(spec)
+
+    already_created_keys = set()
+    if not is_creation:
+        # Here we get the pre-existing tuples and assign them the minimal rank
+        for dependencies in tuple_graph.values():
+            for dependency_key in dependencies:
+                if dependency_key not in tuple_graph:
+                    already_created_keys.add(dependency_key)
```
jmorel, line 92: Could you rename key in id here?

substra/sdk/compute_plan.py
```diff
@@ -0,0 +1,157 @@
...
+
+    # Compute the relative ranks of the new tuples (relatively to each other, these
+    # are not their actual ranks in the compute plan)
+    visited = graph.compute_ranks(
```
jmorel, line 96: Could you rename this variable "ranks"?

substra/sdk/compute_plan.py
```diff
@@ -0,0 +1,157 @@
...
+                tmp_spec.composite_traintuples,
+                tmp_spec.testtuples,
+            ) = graph.filter_tuples_in_list(
+                sorted_by_rank[start:end],
+                traintuples,
+                aggregatetuples,
+                composite_traintuples,
+                testtuples,
+            )
+            yield tmp_spec
```
jmorel, line 134: A generator, nice!

substra/sdk/compute_plan.py
```diff
@@ -0,0 +1,157 @@
...
+    """Auto batching of the compute plan tuples
+    """
+
+    # Create the dependency graph and get the dict
+    # of tuples by id
+    (
+        tuple_graph,
+        traintuples,
+        aggregatetuples,
+        composite_traintuples,
```
jmorel, line 83: This is just a naming convention, but I feel that naming these variables traintuples_idx would make it clearer that they are dict which keys are ids. Just naming them `traintuples` looks like a list.

jmorel (commented):
Thanks a lot for the refactoring and for the explanations, I can finally say I'm confident in how the code works :)

substra/sdk/compute_plan.py
```diff
@@ -101,8 +101,9 @@ def auto_batching(spec, is_creation: bool = True, batch_size: int = 20):
     testtuples = dict()
     if spec.testtuples:
         for testtuple in spec.testtuples:
+            # Rank 0 if testtuple.traintuple_id is in the nodes to ignore
             visited["test_" + testtuple.traintuple_id] = (
-                visited[testtuple.traintuple_id] + 1
+                visited.get(testtuple.traintuple_id, -1) + 1
```
jmorel: This is clever but it could be hard to maintain in the future, maybe just write an if-else clause ?

substra/sdk/graph.py
```diff
@@ -67,11 +69,9 @@ def compute_ranks(
     ranks = ranks or dict()
     visited = set()
     node_to_ignore = node_to_ignore or set()
-    inverted_node_graph = _get_inverted_node_graph(node_graph, node_to_ignore)
 
-    # Number of nodes to visit
-    # TODO check whether this is necessary
-    n_nodes = len(node_graph.keys())  # len(set(node_graph.keys()).difference(node_to_ignore))
+    assert len(set(node_graph.keys()).intersection(node_to_ignore)) == 0
```
jmorel, line 73: Could you rework this assertion into an if/raise ?

substra/sdk/graph.py
```diff
@@ -82,15 +82,15 @@ def compute_ranks(
...
         visited.add(current_node)
         for child in inverted_node_graph.get(current_node, list()):
             ranks[child] = max(ranks[current_node] + 1, ranks.get(child, -1))
 
             # Cycle detection
             edge = (current_node, child)
-            if (edge[1], edge[0]) in edges:  # if edge in edges or
+            if (edge[1], edge[0]) in edges:
                 raise exceptions.InvalidRequest(
                     "missing dependency among inModels IDs", 400
```
jmorel, line 95: We could use a very specific message here, maybe `Circular dependency among tuples between {edge[0]} and edge[1]`

jmorel (changes requested):

substra/sdk/backends/remote/backend.py
Esadruhn: Yes, I'll do it. I want to point out that this is an error for the dev though, not the user

substra/sdk/graph.py
Esadruhn: My problem is that in substra-tests, this test: `test_compute_plan_circular_dependency_failure` asserts that the error message is the same as the one returned by the backend

```python
with pytest.raises(substra.exceptions.InvalidRequest) as e:
        client.add_compute_plan(cp_spec)

assert 'missing dependency among inModels IDs' in str(e.value)
```

so I can add the edges at fault to the message but I need to keep the rest of the exception as is or change the test. I would like to keep something close to what the backend returns too (maybe we should return a clearer message from the backend, this one seems to refer to a different problem).

substra/sdk/compute_plan.py
Esadruhn: I try to take advantage of them more, it seemed like a good occasion !

substra/sdk/backends/local/dal.py
```diff
@@ -109,6 +109,7 @@ def get_with_files(self, type_: schemas.Type, key: str):
             if not tmp_directory.exists():
                 pathlib.Path.mkdir(tmp_directory)
 
+            if not asset_path.exists():
```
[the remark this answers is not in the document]
Esadruhn: Indeed, I think I realised that if we ever had an asset for which we wanted to save several files, then only the first one would be saved. I will move it to another PR

substra/sdk/graph.py
jmorel: Agreed, 👍

jmorel (approved):
Very nice! Let's not merge until the companion PR is validated though.

substra/sdk/backends/remote/backend.py
```diff
@@ -109,26 +113,103 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+            else:
+                with tmp_spec.build_request_kwargs(**spec_options) as (data, files):
+                    asset = self._add(
+                        schemas.Type.ComputePlan,
+                        data,
+                        files=files,
+                        exist_ok=exist_ok
+                    )
+                compute_plan_id = asset['computePlanID']
+            id_to_keys.update(asset['IDToKey'])
```
samlesu, line 176: I think it would be more readable to move out of the for loop the creation method:

```
first_batch, update_batches = ...
create_cp(first_batch)
for batch in update_batches:
    update_cp(batch)
```

substra/sdk/backends/remote/backend.py
```diff
@@ -109,26 +113,103 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+
+        # Special case: no tuples
```
samlesu, line 178: What's this special case? It seems adding a compute plan without tuples should not be supported.

substra/sdk/backends/remote/backend.py
```diff
@@ -109,26 +113,103 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
+                    spec=spec,
+                    spec_options=spec_options
+                )
+            else:
+                return self.add(spec=spec, exist_ok=exist_ok, spec_options=spec_options)
+
+        asset['IDToKey'] = id_to_keys
+        return asset
+
+    def update_compute_plan(self, compute_plan_id, spec, spec_options=None):
```
samlesu, line 194: I don't think we should support the autobatching for the update_compute_plan method.

This method has been created to support mannually the autobatching and to be able to create big compute plans. Updating existing compute plans is not supported by the platform as we cannot update a done compute plan.

In my opinion, we should warn the user in the method documentation (or say it's deprecated) and tell him to prefer the `create_compute_plan()` method.

substra/sdk/backends/remote/backend.py
Esadruhn: It is linked to the `update_compute_plan` function I think, today a DS can create an empty CP then do a loop to add the tuples by calling `update_compute_plan`.  
It is also a test in substra,`test_update_compute_plan`, but that can be updated.

substra/sdk/backends/remote/backend.py
Esadruhn: For the case where users would prefer adding everything at once, it is redundant indeed.  
Weren't there cases where the user did not know when to stop the compute plan (one tuple = one epoch of training for example) and they would update it continuously ?

substra/sdk/backends/remote/backend.py
Esadruhn: Updated it. I also changed the function so that instead of calling self._client.request themselves, they call the `add` and `update_compute_plan` methods without autobatching

substra/sdk/backends/remote/backend.py
jmorel: @Esadruhn is absolutely right here :) This allows allows to create a CP with tag and other attributes and then add tuples one by one to the compute plan. Creating a new CP through a first tuple with rank 0 doesn't allow to have CP attributes set.

samlesu: Thanks for the update. Maybe adding a bit more context would be helpful for future developers? 

```suggestion
        # Special case: no tuples; this is needed when doing a manual auto batching
```

substra/sdk/backends/remote/backend.py
samlesu: Yes, you're right. It makes sense.

This use case is not fully supported by the platform (local folders deleted when compute plan is done by default) but this is an other story and should be treated separately.

samlesu (approved):

PR #218, opened 2020-08-31
feat: add options for auto batching in the CLI

Esadruhn:
In the CLI, when
- adding a compute plan
- updating a compute plan

add the following options:
- disable the auto batching
- set the batch size

### DIFF
--- a/references/cli.md
+++ b/references/cli.md
@@ -283,7 +283,13 @@ Usage: substra add compute_plan [OPTIONS] PATH
       "metadata": dict
   }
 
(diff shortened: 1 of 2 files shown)


jmorel (approved):

jmorel (changes requested):
The documentation says that the default batch size is 20, could you add it as the option default so that it shows up in the doc when doing --help ?

jmorel (dismissed):

jmorel (approved):

PR #220, opened 2020-09-03
Use key instead of pkhash

Kelvin-M:
The goal of this multi-repo PRs is to make sure that all the asset creation endpoints returns only the "key" of the asset. 
Indeed, currently `create` and 'get' an asset return, respectively, `pkhash` and `key`. As `pkhash` is only a substra-backend concept, we need to move from `pkhash` to `key` when dealing with substra cli/sdk and frontend.

Consequently, when client create a new asset, the substra-backend will now only return the `key` of the created asset.

There is an exception for compute plan which need to be returned in its complete form because client need to get the ID_to_key mapping and it’s returned only when client create this asset.

The client will take care of doing an extra GET in order to retrieve the complete asset.

PRs :
    https://github.com/SubstraFoundation/substra-backend/pull/312
    https://github.com/SubstraFoundation/substra/pull/220
    https://github.com/SubstraFoundation/substra-tests/pull/140

### DIFF
--- a/substra/sdk/backends/remote/rest_client.py
+++ b/substra/sdk/backends/remote/rest_client.py
@@ -212,7 +212,7 @@ class Client():
             if not exist_ok:
                 raise
 
-            key = e.pkhash or e.computePlanID
+            key = e.key or e.computePlanID
             is_many = isinstance(key, list)
             if is_many:
                 logger.warning("Many assets not compatible with 'exist_ok' option")
@@ -233,7 +233,7 @@ class Client():
             return self._add(name, exist_ok=exist_ok, **request_kwargs)
 
         except exceptions.RequestTimeout as e:
-            key = e.pkhash
+            key = e.key
             is_many = isinstance(key, list)  # timeout on many objects is not handled
             if not retry_timeout or is_many:
                 raise e
--- a/substra/sdk/backends/remote/rest_client.py
+++ b/substra/sdk/backends/remote/rest_client.py
@@ -212,7 +212,7 @@ class Client():
             if not exist_ok:
                 raise
 
-            key = e.key or e.computePlanID
+            key = e.key
             is_many = isinstance(key, list)
             if is_many:
                 logger.warning("Many assets not compatible with 'exist_ok' option")
--- a/examples/compute_plan/scripts/add_compute_plan.py
+++ b/examples/compute_plan/scripts/add_compute_plan.py
@@ -48,7 +48,7 @@ algo_key = client.add_algo({
         'public': False,
         'authorized_ids': [],
     }
-}, exist_ok=True)['pkhash']
+}, exist_ok=True)['key']
 
 
 print('Generating compute plan...')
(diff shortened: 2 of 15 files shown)


natct10:
+ [documentation](https://github.com/SubstraFoundation/substra-documentation/pull/110)

substra/sdk/backends/remote/rest_client.py
```diff
@@ -212,7 +212,7 @@ def _add(self, name, exist_ok=False, **request_kwargs):
             if not exist_ok:
                 raise
 
-            key = e.pkhash or e.computePlanID
+            key = e.key or e.computePlanID
```
Esadruhn, line 215: actually e does not have a 'computePlanID' attribute, you can directly use `e.key`

Kelvin-M: You're right (https://github.com/SubstraFoundation/substra/blob/bb0cf81b237d6e3f20a0dda034141193fa4db20d/substra/sdk/exceptions.py#L113-L117)

Esadruhn:
https://github.com/SubstraFoundation/substra/pull/220/files#diff-434b67ddd91e180773c318731c47bcfdR138

Is removing the call to `get` after the `add` part of this MR ? Or is it OK if I commit on this MR ?

Esadruhn:
> > https://github.com/SubstraFoundation/substra/pull/220/files#diff-434b67ddd91e180773c318731c47bcfdR138
> > Is removing the call to `get` after the `add` part of this MR ? Or is it OK if I commit on this MR ?
> 
> @Esadruhn You can commit in the PR for this part 👍

It is a lot of changes in the end (with the choice of getting either the key or asset) so I put it in a separate MR: #221

Esadruhn (approved):
LGTM !

substra/sdk/backends/remote/rest_client.py
```diff
@@ -212,7 +212,7 @@ def _add(self, name, exist_ok=False, **request_kwargs):
             if not exist_ok:
                 raise
 
-            key = e.pkhash or e.computePlanID
+            key = e.key
```
jmorel, line 215: Are you sure you can remove handling the computePlanID here? Are we sure there cannot be an "AlreadyExists" exception when creating a compute plan?

substra/sdk/backends/remote/rest_client.py
Kelvin-M: @jmorel

substra/sdk/backends/remote/rest_client.py
Kelvin-M: See there : https://github.com/SubstraFoundation/substra/pull/220#discussion_r482822939

Kelvin-M:
Need review on this one : @jmorel @inalgnu @samlesu

jmorel (approved):

inalgnu (approved):

PR #221, opened 2020-09-03
substra client backend: do not do a systematic 'get' after creating an asset

Esadruhn:
With this PR: https://github.com/SubstraFoundation/substra/pull/220 the backend returns only a key instead of the whole object (except for the compute plan). 

The SDK client function 'add' returns this key, it is the responsibility of the user to do a `get` if he wants the whole object. To help him with possible timeouts, we expose the `utils.retry_on_exception` function.
For the compute plan, it always returns the whole object, otherwise we lose the id_to_key mapping.

## Changes

- the backend returns only the key of the asset after a 'add' (except for the compute plan)
- substra SDK:
  - substra only returns the key of the asset
  - the function 'retry_on_exception' is exposed to the user
- substra CLI:
  - the CLI returns the whole object after a 'add' (using the `retry_on_exception` function)
  - there is a new command line argument `timeout` to change the timeout of the `get` request after an `add`
- substra SDK, local backend:
  - same behaviour as the remote backend
  - fixed a bug: replaced `+ keys or list()` by `+ (keys or list())`

## Check list

- [x] Documentation
- [x] substra tests pass
- [x] substra-tests tests pass
- [x] substra-tests local tests pass

## Companion PR
https://github.com/SubstraFoundation/substra-tests/pull/141


Kelvin-M (approved):

samlesu:
I think the name of the option could be improved. Maybe `return_key_only` / `return_asset` would be more explicit?

It may be confusing to return different responses depending on input arguments. I would be in favour of either always returning a dict (with a single key field) or letting the user do an explicit call to `get_` to get the object and remove this option.

Kelvin-M (approved):

AurelienGasser:
I agree with @samlesu 's second suggestion : doing an explicit `get` after each `add` seems cleaner overall.

references/sdk.md
```diff
@@ -69,7 +69,7 @@ Returns:
 
 ## add_data_sample
 ```python
-Client.add_data_sample(self, data, local=True, exist_ok=False)
+Client.add_data_sample(self, data, local=True, exist_ok=False, get_asset=False)
```
jmorel, line 72: There is not get_data_sample if I'm not mistaken.

references/sdk.md
```diff
@@ -100,10 +100,12 @@ raised.
 If `exist_ok` is true, `AlreadyExists` exceptions will be ignored and the
 existing asset will be returned.
 
+If `get_asset` is False, only the key is returned, otherwise the object is returned.
+
 
 ## add_data_samples
 ```python
-Client.add_data_samples(self, data, local=True)
+Client.add_data_samples(self, data, local=True, get_asset=False)
```
jmorel, line 108: Same as for "add_data_sample"

substra/sdk/backends/remote/backend.py
```diff
@@ -127,15 +127,16 @@ def add(self, spec, exist_ok=False, spec_options=None):
...
-        # Do not do this with a compute plan or we lose the id_to_key field
+        # The backend returns only the key, except for the compute plan: it returns the
+        # whole object (otherwise we lose the id_to_key field)
         if asset_type == schemas.Type.ComputePlan:
             return response
 
-        key = _get_asset_key(response)
-        return self.get(asset_type, key)
+        if get_asset:
+            key = _get_asset_key(response)
```
jmorel, line 136: This is the only use of _get_asset_key and here it can never be a compute plan (handled a couple of lines above). You could probably remove the function and just do `return self.get(asset_type, reponse['key']`

samlesu:
If we want to let the user ensure the asset has been created, I think we should have an option `wait_until_created`. In my opinion, in both cases the function should return a dict object representing the asset. 

Could we have some issues if we don't wait for the asset(s) to be created? For instance we won't be able to link a datasample with a "pending" dataset? Or when creating a tuple depending on a "pending" algo? 
If yes, we should set this option to True by default as it would be safer (and we would be backward compatible).

Kelvin-M:
> If we want to let the user ensure the asset has been created, I think we should have an option `wait_until_created`. In my opinion, in both cases the function should return a dict object representing the asset.

It's the `get_asset` option with the retry mechanism (`utils.retry_on_exception`) which does that, no ? Are you talking about a name issue or functionnal issue ?
@Esadruhn seems to explain it in the previous post.

Esadruhn:
Decision with Claire, Kelvin and Jérémy:
- the 'add' method always returns the key (except for `add_compute_plan`)
- the 'get' method has an extra parameter `wait_until_created` to retry on exception

samlesu:
> > If we want to let the user ensure the asset has been created, I think we should have an option `wait_until_created`. In my opinion, in both cases the function should return a dict object representing the asset.
> 
> It's the `get_asset` option with the retry mechanism (`utils.retry_on_exception`) which does that, no ? Are you talking about a name issue or functionnal issue ?
> @Esadruhn seems to explain it in the previous post.

Yes, good point, my bad! It seems like timeouts are always handled by the add method, even if the `get_asset` option is set to False. The name `wait_until_created` does not make sense. The timeouts are handled using a retry on the add method; so that's perfect, with or without timeout the return value will be consistent. So there is no functional issue.

references/cli.md
```diff
@@ -49,6 +49,9 @@ Usage: substra login [OPTIONS]
   Login to the Substra platform.
 
 Options:
+  --timeout INTEGER               Set timeout (seconds) for retrying a call
+                                  [default: pretty]
```
Kelvin-M, line 53: What does pretty, for the default value, mean here ?

Esadruhn: it's a typo, should be a float

Esadruhn:
Updated the description with the latests changes

substra/sdk/utils.py
```diff
@@ -158,7 +158,22 @@ def parse_filters(filters):
...
+    from substra.sdk import exceptions, retry_on_exception
+
+    def my_function(arg1, arg2):
+        pass
+
+    retry = retry_on_exception(
+                exceptions=(exceptions.RequestTimeout),
+                timeout=300,
+            )
+    retry(my_function)(arg1, arg2)
```
jmorel, line 175: it looks like you're missing the closing ` ``` `

jmorel (dismissed):

Esadruhn:
@AurelienGasser @samlesu is the last version OK with you ?

Kelvin-M (dismissed):

jmorel (dismissed):

substra/sdk/utils.py
```diff
@@ -158,7 +158,23 @@ def parse_filters(filters):
 
 
 def retry_on_exception(exceptions, timeout=300):
-    """Retry function in case of exception(s)."""
+    """Retry function in case of exception(s).
```
AurelienGasser, line 161: I think it'd be worth specifying that the `timeout` is in seconds

substra/cli/interface.py
```diff
@@ -189,6 +198,19 @@ def click_option_expand(f):
...
+def click_global_conf_retry_timeout(f):
+    """Add timeout option to command."""
+    return click.option(
+        '--timeout', 'timeout',
+        type=click.INT,
+        expose_value=False,
+        default=DEFAULT_RETRY_TIMEOUT,
+        show_default=True,
+        callback=update_global_conf,
+        help='Set timeout (seconds) for retrying a call'
```
AurelienGasser, line 210: This help text confused me a bit (I was not sure if this was actually a "delay" between each call). 

I think it would be more clear with:

```suggestion
        help='Max number of seconds the operation will be retried for'
```

What do you think?

AurelienGasser (changes requested):
Some minor feedback otherwise LGTM

substra/cli/interface.py
Esadruhn: Yes it is clearer !

Kelvin-M (approved):

jmorel (approved):

Issue #200, opened 2020-07-09
[Feature request] Save the authentication token when logging from the SDK


Esadruhn:
## Current behaviour

When the user logs in from the CLI, the token is saved so he can create a client from the SDK (with the same configuration) and use the API without logging again.

When the user logs in from the SDK, the token is not saved so he has to log again if he uses the CLI. It also means that if the user is logged in the CLI and logs in the SDK, a new token is generated and he is no longer logged in the CLI.

## Feature

Automatically save the authentication token when logging from the SDK.
It is possible only if the user logs from a config file (there is no profile name otherwise, nor config_path, token_path)

### Caveat

This means that if the user logs from a config file, the token is saved, but if he creates the client (`__init__` function) then uses the `login` function to log in, it does not work.
A workaround would be to create a `save_token` function (and/or `save_profile`) which asks for a profile name and config_paths/token_paths and saves the token / the profile.

natct10:
Hi, this would really ease end users workflow using both SDK & CLI without re-login/new token!

samlesu:
If there is no profile name, and no config file, we could update the default profile and the default config file?

I'm not in favour of adding new methods to manually handle token and profile, I think it's better if it's handled under the hood and the user does not have to deal with this.

jmorel:
> If there is no profile name, and no config file, we could update the default profile and the default config file?

This could erase valid tokens. I'd say that we only save the token if the client was instantiated via "from_config_file". Because if not, we do not know with certainty where to save the token and for which profile.

Esadruhn:
Thank you for your feedback !

Actually I realise either the code has changed or I read it badly (probably the case, sorry 😔 ), the behaviour is not exactly what I wrote.

For now the user can connect in two ways with the SDK:

1. Using the `__init__` and `login` functions 

```python
client = Client(url= ..., .., token=...)
# If we do not give a valid token as input, need to login
client.login(username=..., password=...) 
```

2. Using the `from_config_file` and `login` functions:

```python
client = Client.from_config_file(profile_name= ..., tokens_path=...., token=....)
# If we do not give a valid token as input, 
# or there is no token in the 'token file' for this profile, need to login
client.login(username=..., password=...) 
```

In both methods, since we do not save the profile name, we do not know which profile the token goes with, with the 
afore mentioned problems of default value and overwriting.

So either we 
- save the profile name and token file path in the client when using 'from_config_file'
- use these values to save the token in the 'login' method
and
[...]

PR #226, opened 2020-09-22
feat: login from SDK in examples and expose temp dir in debug mode

Esadruhn:
User feedback:

- Login from the SDK in the examples, instead of logging in from the CLI then use the SDK
- Expose the temporary directory in debug mode so that the user an pause the execution and check the assets

EDIT: only expose the temp dir following the discussions on the login

### DIFF
--- a/references/sdk.md
+++ b/references/sdk.md
@@ -30,6 +30,10 @@ Args:
         the deployed Substra platform. The platform is in read-only mode.
         Defaults to False.
 
+## temp_directory
+Temporary directory for storing assets in debug mode.
+Deleted when the client is deleted.
+
 ## login
 ```python
 Client.login(self, username, password)
--- a/substra/sdk/backends/local/backend.py
+++ b/substra/sdk/backends/local/backend.py
@@ -32,6 +32,12 @@ class Local(base.BaseBackend):
         self._db = dal.DataAccess(backend)
         self._worker = compute.Worker(self._db)
 
+    @property
+    def temp_directory(self):
+        """Get the temporary directory where the assets are saved.
+        The directory is deleted at the end of the execution."""
+        return self._db.tmp_dir
+
     def login(self, username, password):
         self._db.login(username, password)
 
--- a/substra/sdk/backends/local/dal.py
+++ b/substra/sdk/backends/local/dal.py
@@ -34,11 +34,11 @@ class DataAccess:
     def __init__(self, remote_backend: typing.Optional[backend.Remote]):
         self._db = db.get()
         self._remote = remote_backend
-        self.__tmp_dir = tempfile.TemporaryDirectory(prefix="/tmp/")
+        self._tmp_dir = tempfile.TemporaryDirectory(prefix="/tmp/")
 
     @property
-    def _tmp_dir(self):
-        return pathlib.Path(self.__tmp_dir.name)
+    def tmp_dir(self):
+        return pathlib.Path(self._tmp_dir.name)
 
     @staticmethod
     def is_local(key: str):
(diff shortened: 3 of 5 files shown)


samlesu:
Storing password in source code is a bad practise. IMHO, we should value security vs usability in our examples. 

If we want to keep using the login method, we could maybe use env variables?

Esadruhn:
Do you mean that this sets a bad example for the user ?

The username and password in themselves were in the README and work with the local install so I don't think there is a problem exposing them.

I can change to an env variable for the good practice

examples/compute_plan/scripts/add_compute_plan.py
```diff
@@ -24,6 +24,7 @@
 compute_plan_keys_path = os.path.join(current_directory, '../compute_plan_keys.json')
 
 client = substra.Client.from_config_file(profile_name="node-1")
+client.login(username="node-1", password="p@$swr0d44")
```
jmorel: We originally had the login call in the examples but every login call invalidate the previous token. As a result, you would have to login using the CLI after each script.

This is why we decided to include the login command in the readme as a prerequisite.

examples/cross_val/scripts/cross_val_algo_random_forest.py
```diff
@@ -65,5 +66,5 @@
 
 print(f'Tuples keys have been saved to {os.path.abspath(folds_keys_path)}')
 print('\nRun the following commands to track the status of the tuples:')
-print(f'    substra list traintuple -f "traintuple:tag:{tag}" --profile node-1')
-print(f'    substra list testtuple -f "testtuple:tag:{tag}" --profile node-1')
+print(f'    substra list traintuple -f "traintuple:tag:{tag}"')
```
jmorel, line 69: I think the profile is necessary here, otherwise calls will use the default profile and fail if the default profile isn't the same as node-1.

examples/titanic/README.md
```diff
@@ -15,8 +15,8 @@ In order to run this example, you'll need to:
 * [pull the `substra-tools` docker images](https://github.com/substrafoundation/substra-tools#pull-from-private-docker-registry)
 * create a substra profile to define the substra network to target, for instance:
     ```sh
-    substra config --profile node-1 http://substra-backend.node-1.com
-    substra login --profile node-1 --username node-1 --password 'p@$swr0d44'
+    substra config --profile default http://substra-backend.node-1.com
+    substra login --profile default --username node-1 --password 'p@$swr0d44'
```
jmorel: If you switch to default you can just remove the --profile option.

jmorel (changes requested):

examples/compute_plan/scripts/add_compute_plan.py
Esadruhn: Yes I understand, and I cheated a bit here since the SDK login is on profile 'node-1' while the CLI login is on profile 'default' so they do not invalidate each other.

I think this is not ideal, but as we discussed it in https://github.com/SubstraFoundation/substra/issues/200 it is not simple to solve.

examples/cross_val/scripts/cross_val_algo_random_forest.py
Esadruhn: The 'default' and 'node-1' profiles are actually equal, one is used to login with the CLI and the other with the SDK, so there is no token conflict.

examples/compute_plan/scripts/add_compute_plan.py
jmorel: Actually they do, because they are for the same username. The token is invalidated in the backend, not just in the local profile.

Esadruhn:
For the history: what I did works, checked with Jérémy and it is because there is "reuse token" strategy on locally deployed instance, so it would not work in production.  
After discussion on why the login was that way in the examples, I revert it back.

jmorel (approved):

PR #228, opened 2020-09-23
Generate the documentation of the SDK - arguments as markdown instead of block quotes

Esadruhn:
The previous way of generating the SDK doc meant that we had to change from the default Google docstring.

For example, this:
```python
def my_func():
       """Returns a new Client configured with profile data from configuration files.

        Args:
            profile_name (str, optional): Name of the profile to load.
                Defaults to 'default'.
            retry_timeout (int, optional): Number of seconds before attempting a retry call in case
                of timeout. Defaults to 5 minutes.

        Returns:
            Client: The new client.
        """
```

Gave something like this: 

```
Returns a new Client configured with profile data from configuration files.

Args:
    profile_name (str, optional): Name of the profile to load. Defaults to 'default'. retry_timeout (int, optional): Number of seconds before attempting a retry call in case of timeout. Defaults to 5 minutes.

Returns: Client: The new client.
```

so we had to insert newlines in the docstring: 
```python
def my_func():
       """Returns a new Client configured with profile data from configuration files.

        Args:

            profile_name (str, optional): Name of the profile to load.
                Defaults to 'default'.

            retry_timeout (int, optional): Number of seconds before attempting a retry call in case
                of timeout. Defaults to 5 minutes.

        Returns:

            Client: The new client.
        """
```

Furthermore, any link inserted in the docstring, like `it must follow the [DataSampleSpec schema](sdk_schemas.md#DataSampleSpec).` would not be rendered.

So, to stop having to change the docstring for the markdown, I updated the way the markdown was generated for the SDK. Instead of using pydocmd, it uses a manual script with the [docstring-parser](https://pypi.org/project/docstring-parser/) library to parse the docstring and display the arguments as a list. 

## What does it fix 

This fixes the links in the doc, for example: https://github.com/SubstraFoundation/substra/blob/master/references/sdk.md#add_data_samples
`data (Union[dict, schemas.DataSampleSpec]): data samples to add. If it is a dict,
    it must follow the [DataSampleSpec schema](sdk_schemas.md#DataSampleSpec).`

And the annoying newlines that we had to add in the Client method docstrings.

### DIFF
--- a/bin/generate_sdk_documentation.py
+++ b/bin/generate_sdk_documentation.py
@@ -52,7 +52,7 @@ def generate_function_help(fh, asset):
         fh.write(f"\n**{param.args[0].title()}:**\n")
         if len(param.args) > 1:
             for extra_param in param.args[1:]:
-                fh.write(f"\n - {extra_param}: ")
+                fh.write(f"\n - `{extra_param}`: ")
         fh.write(f"{param.description}\n")
 
 
--- a/bin/generate_sdk_documentation.py
+++ b/bin/generate_sdk_documentation.py
@@ -38,7 +38,7 @@ def generate_function_help(fh, asset):
         fh.write(f"{docstring.long_description}\n")
     # Write the arguments as a list
     if len(docstring.params) > 0:
-        fh.write("\nArguments:\n")
+        fh.write("\n**Arguments:**\n")
     for param in docstring.params:
         type_and_optional = ""
         if param.type_name or param.is_optional is not None:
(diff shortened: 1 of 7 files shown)


jmorel:
I'm loving this! This could allow for a smooth transition to proper google-style docstrings, something we already attempted to do but in the end didn't pursue because the generate doc was so bad.

bin/generate_sdk_documentation.py
```diff
@@ -0,0 +1,91 @@
...
+    if len(docstring.params) > 0:
+        fh.write("\nArguments:\n")
+    for param in docstring.params:
+        type_and_optional = ""
+        if param.type_name or param.is_optional is not None:
+            text_optional = 'required'
+            if param.is_optional:
+                text_optional = 'optional'
+            type_and_optional = f"({param.type_name}, {text_optional})"
+        fh.write(f" - {param.arg_name}{type_and_optional}: {param.description}\n")
```
jmorel, line 49: Could you make the arg_name and type_name as code (using backquotes) and could you add a space between the arg_name and the type_and_optional? I feel it would improve the readability a bit.

bin/generate_sdk_documentation.py
```diff
@@ -0,0 +1,91 @@
...
+    fh.write("\n\n")
+    # Write the docstring
+    docstring = inspect.getdoc(asset)
+    docstring = docstring_parser.parse(inspect.getdoc(asset))
+    fh.write(f"{docstring.short_description}\n")
+    if docstring.long_description:
+        fh.write(f"{docstring.long_description}\n")
+    # Write the arguments as a list
+    if len(docstring.params) > 0:
+        fh.write("\nArguments:\n")
```
jmorel, line 41: Could you make arguments and returns bold? Making them into level3 headers may be too much, but a bit of bold would be nice:)

jmorel (approved):

PR #235, opened 2020-10-20
Use random UUIDs for asset keys (instead of content hashes)

AurelienGasser:
Companion PRs:

- https://github.com/SubstraFoundation/substra-tests/pull/153
- https://github.com/SubstraFoundation/substra-backend/pull/342
- https://github.com/SubstraFoundation/substra-chaincode/pull/128

**Please see the main PR in substra-backend for a detailed description of the changes:** https://github.com/SubstraFoundation/substra-backend/pull/342

---

The backend changes have been reflected in the local backend and the SDK:

- Keys are now UUIDs and not hashes of the asset content
- It is now allowed to register the same asset twice (drop support for `exist_ok`, remove most `AlreadyExists` exceptions)

Note that `409 Conflict` (aka AlreadyExists) errors can still be returned by the backend in a specific edge cases, so they are still supported in the SDK code. The **only** case that can trigger a `409 Conflict` error is when the backend cannot receive a response from the blockchain because of network instability. In that case, the backend will retry its request to the blockchain: if the first request had been honored correctly, then the second request will fail (`AlreadyExists`) and that error will be propagated to the SDK (`409 Conflict`). The error handling mechanism in that case (since we received a `409` answer instead of the full asset details) is to issue an extra GET request to the backend to fetch the asset details. This mechanism already existed and is unchanged.

# Future work

To simplify diffs in this PR, some work has been postponed to a future PR (which should happens shortly after this PR is merged):

## Rename fields

To simplify APIs and make them consistent, rename the following fields

- `dataset.keys` => `dataset.data_sample_keys`
- `compute_plan_id` => `key` / `compute_plan_key`
- `.*hash` => `.*checksum`


Kelvin-M (approved):

Esadruhn (commented):
Thanks for the PR ! Only a few remarks, all good otherwise (I still have to test, will do it ASAP)

substra/sdk/backends/remote/rest_client.py
```diff
@@ -204,28 +204,23 @@ def _add(self, name, exist_ok=False, **request_kwargs):
...
             key = e.key
             is_many = isinstance(key, list)
             if is_many:
-                logger.warning("Many assets not compatible with 'exist_ok' option")
+                logger.warning("AlreadyExists exception was received for a list of keys. "
+                               "Unable to determine which key(s) already exist.")
                 raise
 
             logger.warning(f"{name} already exists: key='{key}'")
             return self.get(name, key)
```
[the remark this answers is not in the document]
AurelienGasser: fixed in https://github.com/SubstraFoundation/substra/pull/235/commits/d8bdb1f15bbb09ca6d3514013a421edc842422ba

substra/sdk/schemas.py
```diff
@@ -142,7 +142,7 @@ def build_request_kwargs(self, local):
             yield data, None
 
     def compute_key(self) -> str:
-        return fs.hash_directory(self.path)
+        return str(uuid.uuid4())
```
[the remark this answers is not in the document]
AurelienGasser: fixed in https://github.com/SubstraFoundation/substra/pull/235/commits/d8bdb1f15bbb09ca6d3514013a421edc842422ba

AurelienGasser:
@Esadruhn Thanks for the review!

substra/sdk/backends/local/backend.py
```diff
@@ -559,12 +555,14 @@ def _add_traintuple(self, key, spec, spec_options=None):
...
-                "hash": spec.algo_key,
+                "key": spec.algo_key,
+                "hash": algo.content.hash_,
                 "name": algo.name,
                 "storage_address": algo.content.storage_address
             },
             dataset={
-                "opener_hash": spec.data_manager_key,
+                "key": spec.data_manager_key,
+                "opener_hash": spec.data_manager_key,  # TODO: pass actual opener hash
```
Esadruhn, line 565: Above in the function we get `data_manager = self._db.get(schemas.Type.Dataset, spec.data_manager_key)` so you can replace by `data_manager.opener.hash_`

substra/sdk/backends/local/backend.py
```diff
@@ -663,15 +662,16 @@ def _add_testtuple(self, key, spec, spec_options=None):
...
                 "metrics": objective.metrics
             },
             traintuple_key=spec.traintuple_key,
             traintuple_type=traintuple.type_.value,
             algo=traintuple.algo,
             certified=certified,
             dataset={
-                "opener_hash": dataset_opener,
+                "key": dataset_key,
+                "opener_hash": dataset_key,  # TODO: pass actual opener hash
```
Esadruhn, line 674: You can get it like this:

```python
dataset = self._db.get(schemas.Type.Dataset, dataset_key)
opener_hash = dataset.opener.hash_
```

substra/sdk/backends/local/backend.py
```diff
@@ -757,12 +759,14 @@ def _add_composite_traintuple(
...
-                "hash": spec.algo_key,
+                "key": spec.algo_key,
+                "hash": algo.content.hash_,
                 "name": algo.name,
                 "storage_address": algo.content.storage_address
             },
             dataset={
-                "opener_hash": spec.data_manager_key,
+                "key": spec.data_manager_key,
+                "opener_hash": spec.data_manager_key,  # TODO: pass actual opener hash
```
Esadruhn, line 769: l.700, we get the dataset for existence validation `self._db.get(schemas.Type.Dataset, spec.data_manager_key)`,
so you can replace it by `dataset = self._db.get(schemas.Type.Dataset, spec.data_manager_key)` and do for the opener hash: 
```python
"opener_hash": dataset.opener.hash_
```

substra/sdk/backends/local/db.py
```diff
@@ -36,11 +36,8 @@ def add(self, asset):
...
 
-        if key in self._data[type_]:
-            raise exceptions.AlreadyExists(key, 409)
-        else:
-            self._data[type_][key] = asset
-            logger.info(f"{type_} with key '{key}' has been created.")
+        self._data[type_][key] = asset
+        logger.info(f"{type_} with key '{key}' has been created.")
 
         return asset
```
Esadruhn, line 43: this about the "get" function below, the "log" argument is now useless (it was to test the existence before adding without logging an error "already exists")

Kelvin-M (dismissed):

substra/sdk/schemas.py
```diff
@@ -210,7 +207,7 @@ class ComputePlanSpec(_BaseComputePlanSpec):
 
     @staticmethod
     def compute_key() -> str:
```
Esadruhn, line 209: You can delete this, `ComputePlanSpec` inherits from `_BaseComputePlanSpec` that inherits from `_Spec` so the `compute_key` function is inherited.

substra/sdk/backends/local/db.py
AurelienGasser: The `log argument is still used in the KeyError case

```
    def get(self, type_, key: str, log: bool = True):
        """Return asset."""
        try:
            return self._data[type_][key]
        except KeyError:
            if log:
                logger.error(f"{type_} with key '{key}' not found.")
            raise exceptions.NotFound(f"Wrong pk {key}", 404)
```

do you want me to remove that `logger.error`?

Esadruhn: On the contrary, now that there is no check in the local backend `add` function, we can always do `logger.error` when the asset  does not exist. The `log` parameter to hide the error message is useless.

jmorel:
I understand why there can still be a 409 returned, but I think this should be handled by the backend entirely and the Client shouldn't even include the AlreadyExists error anymore. We're clearly removing from the documentation the fact that there can be conflicts, saying that duplication is now fine. In that context we really shouldn't ever get a 409 returned to the end user.

jmorel (changes requested):
According to the readme, dataset.keys has been renamed in dataset.data_sample_keys, but I don't see the matching rename in printers.py and worker.py

AurelienGasser:
@jmorel 

> I think this should be handled by the backend entirely and the Client shouldn't even include the AlreadyExists error anymore

The 409 / AlreadyExists is not returned to the end user: it is [caught by the REST client](https://github.com/SubstraFoundation/substra/pull/235/files#diff-921b95f4cd4ad6d612fb452660f7a080e90b5f67feda6d6f34213f504874c000R215) and there are two cases:

1. either it's a compute plan and we issue a `GET` on the backend to retrieve the full asset
2. or it's any other type of asset and the `409` contains the offending key, which is all we need to return, so the error is invisible to the client

There could be an argument for making the `GET` in case one directly in the backend. I discussed it with @inalgnu who argued we should try avoid overwhelming the backend with retries (we already do a lot of retries).

jmorel:
Thank you for the explanations, if the technical choice has already been discussed then I'm good.

jmorel (approved):

Esadruhn (approved):

AurelienGasser:
Thanks all for the review

PR #238, opened 2020-11-02
Add support for 'multi nodes' to the debug mode

Esadruhn:
## Description

In "normal" mode, the user connects to a node, adds the data on this node. When a DS creates a traintuple (or another task), it is executed where the data is: on the node of the dataset. To know which node a dataset is on, get the `dataset.owner` attribute.

The goal of this PR is to simulate this behaviour in debug mode.

- **What it does**: 
  - for the [chainkey use case](https://github.com/SubstraFoundation/substra/pull/239/files), get the appropriate chainkey for each tuple. So with this other PR, the public feature 'support chainkeys' is added to the debug mode.
- **What it does not do**: there is no check on the permissions in debug mode, so impossible to test them with this (but they are computed so the user can display them)

## Changes

In debug mode, the user can assign a specific node as the "owner" of the data: 
- when creating a dataset asset, add a special key in the `metadata` field: `substra.DEBUG_OWNER`. The value will be the owner of the dataset.

This MR also implements the management of the "worker" as in the backend:
- the "worker" in `traintuple.dataset.worker`, `composite_traintuple.dataset.worker`, `objective.test_dataset.worker` is equal to the `owner` of the dataset they are linked to.
- as before, the "worker" of an aggregate tuple (`aggregatetuple.worker`) is specified by the user

The default owner is `local-backend`.

### EDIT

Following the PR comments, added the possibility to change the `creator` of all assets with the `substra.DEBUG_OWNER` key in the metadata. This is only for display.

## Example (incomplete code)

```python
client = substra.Client(debug=True)
dataset_query = { ... }  # All the arguments to create a dataset
dataset_query['metadata'] = {substra.sdk.DEBUG_OWNER: 'owner_1'}
dataset_key = client.add_dataset(dataset_query)
asset = client.get_dataset(key)
assert asset.owner == 'owner_1'

traintuple_query = { ... }  # All the arguments to create a traintuple
traintuple_query['data_manager_key'] = dataset_key
traintuple_key = client.add_traintuple(traintuple_query)
traintuple = client.get_traintuple(traintuple_key)
assert traintuple.dataset.worker == 'owner_1'
```

## How to test

### DIFF
--- a/references/sdk.md
+++ b/references/sdk.md
@@ -36,6 +36,8 @@ add_aggregate_algo(self, data: Union[dict, substra.sdk.schemas.AggregateAlgoSpec
 ```
 
 Create new aggregate algo asset.
+In debug mode, add the following key: `substra.DEBUG_OWNER` to the metadata,
+the value becomes the 'creator' of the objective.
 
 **Arguments:**
  - `data (Union[dict, schemas.AggregateAlgoSpec], required)`: If it is a dict,
@@ -51,6 +53,8 @@ add_aggregatetuple(self, data: Union[dict, substra.sdk.schemas.AggregatetupleSpe
 ```
 
 Create a new aggregate tuple asset.
+In debug mode, add the following key: `substra.DEBUG_OWNER` to the metadata,
+the value becomes the 'creator' of the objective.
 
 **Arguments:**
  - `data (Union[dict, schemas.AggregatetupleSpec], required)`: If it is a dict, it must have the same
@@ -66,6 +70,8 @@ add_algo(self, data: Union[dict, substra.sdk.schemas.AlgoSpec]) -> str
 ```
 
 Create new algo asset.
+In debug mode, add the following key: `substra.DEBUG_OWNER` to the metadata,
+the value becomes the 'creator' of the objective.
 
 **Arguments:**
  - `data (Union[dict, schemas.AlgoSpec], required)`: If it is a dict, it must have the same keys
@@ -80,6 +86,8 @@ add_composite_algo(self, data: Union[dict, substra.sdk.schemas.CompositeAlgoSpec
 ```
 
 Create new composite algo asset.
+In debug mode, add the following key: `substra.DEBUG_OWNER` to the metadata,
+the value becomes the 'creator' of the objective.
 
 **Arguments:**
  - `data (Union[dict, schemas.CompositeAlgoSpec], required)`: If it is a dict, it must have the same
@@ -96,6 +104,9 @@ add_composite_traintuple(self, data: Union[dict, substra.sdk.schemas.CompositeTr
 Create new composite traintuple asset.
 As specified in the data structure, output trunk models cannot be made
 public.
+
+In debug mode, add the following key: `substra.DEBUG_OWNER` to the metadata,
+the value becomes the 'creator' of the objective.
 
 **Arguments:**
  - `data (Union[dict, schemas.CompositeTraintupleSpec], required)`: If it is a dict, it must have the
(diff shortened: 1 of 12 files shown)


samlesu:
In my opinion, one the key feature of the debugging mode is to let user test their code by switching on a single option. They have (almost) nothing else to know.

It looks like that this PR is not going in that direction. I'm wondering if it's possible to not introduce new mechanism in the public code to test private feature.

Esadruhn:
In waiting state until it is confirmed that this is needed and that we do it that way.

substra/sdk/__init__.py
```diff
@@ -14,8 +14,10 @@
 
 from substra.sdk.client import Client
 from substra.sdk.utils import retry_on_exception
+from substra.sdk.backends.local.backend import DEBUG_OWNER
 
 __all__ = [
     'Client',
     'retry_on_exception',
+    'DEBUG_OWNER',
```
jmorel, line 22: To facilitate usage, could we also add it to substra/__init__.py so that users can write `substra.DEBUG_OWNER` instead of `substra.sdk.DEBUG_OWNER` ? The SDK is an internal structure that we don't want to expose to users.

references/sdk.md
```diff
@@ -175,6 +175,9 @@ add_dataset(self, data: Union[dict, substra.sdk.schemas.DatasetSpec])
 ```
 
 Create new dataset asset and return its key.
+In debug mode, add the following key:`substra.sdk.DEBUG_OWNER` to the metadata,
+the value becomes the node owner of the data, and all tuples using this data
+have their worker set to this node.
```
jmorel, line 180: Is this only for datasets or shouldn't this mechanism work for all assets?

jmorel (changes requested):

references/sdk.md
Esadruhn: Which assets are you thinking about ?

It is set on the dataset, and then if:
- an objective
- a traintuple
- a composite traintuple
- a testtuple
uses this dataset, then their worker is set to that value, like in the backend.  

The alternative would be to add this to the data samples, notably because if the objective is not linked to the dataset then we cannot get the right worker.  

The problem is that linking it to the datasamples means that whenever we link a dataset to a data sample, we have to check that the worker of the dataset is the same for all its data samples, and a dataset with no data samples (at creation) has no worker. It may be the right way to do it, did not find the equivalent in the backend (it is in the go code, I had trouble understanding how it worked.)

jmorel: I'm sorry I wasn't clear. If I'm not mistaken, the "real" substra put an owner field on all objects. My thinking was that we should also allow to setup this owner field for all objects purely for consistency even though we'll only use the value from the dataset object.

Esadruhn: Done, you can change the 'creator' for all assets (called 'owner' for the dataset) except for the data samples since they have no metadata.

references/sdk.md
```diff
@@ -36,6 +36,8 @@ add_aggregate_algo(self, data: Union[dict, substra.sdk.schemas.AggregateAlgoSpec
 ```
 
 Create new aggregate algo asset.
+In debug mode, add the following key: `substra.DEBUG_OWNER` to the metadata,
+the value becomes the 'creator' of the objective.
```
jmorel, line 40: You've kept `objective` even though this is an aggregate algo. The issue is present pretty much everywhere in the doc.

jmorel (changes requested):

references/sdk.md
Esadruhn: copy paste failure !

jmorel (approved):

jeandut:
As mentioned in this issue it would be nice if we could have the wanted behavior without forcing the user to add any extra field whatismore from my understanding this PR doesn't allow to deal with different compute plans from the same data owner or at least not straightforwardly.

Wouldn't it be possible to just use different local folders indexed by compute_plan_key and possibly data manager key (local_cpa_worker1, local_cpa_worker2, local_cpb_worker1) or am I missing something ?

Esadruhn:
> e wanted behavior without forcing the user to add any extra field whatismore from my understanding this PR doesn't allow to deal with different compute plans fr

The different compute plans were already dealt with in the local-worker folder. With this issue, what is in the local-worker folder:
```
local_worker  
|__ compute_plans  
    |__ {node_id}  
        |__ {compute_plan_key}
            |__ stuff from the shared compute plan volume on this node
|__ performances
    |__ {testtuple_key}
        |__ performance.json 
|__ models
    |__ {model_key}
        |__ model    # saved model as in `save_model` in the algo.py
```

PR #239, opened 2020-11-03
Enable chainkey support in debug mode

Esadruhn:
## Description

Enable chainkey support in debug mode. This needs the 'debug_multi_nodes' feature #238 

## Changes

- init the local backend:
  - env variables "CHAINKEYS_ENABLED" and "CHAINKEYS_DIR". I chose environment variables to avoid having extra variables specific to the debug mode in the Client initialization.
- Worker: get the chainkey that matches the `tuple_.dataset.worker`, and give the env variable "NODE_INDEX" to the Docker container for the task

## Bug fixes

- aggregate tuple that has a composite traintuple in its in models: it took the head instead of the trunk
- Compute plan - get id to rank: always tried to get a traintuple from the db, even if composite traintuple or aggregatetuple
- DB in debug mode: the db was always the same so if you create several clients in the same context, they share the db (caused errors in the tests in melloddy since we check the number of datasets on the client)

### DIFF
--- a/substra/sdk/backends/local/backend.py
+++ b/substra/sdk/backends/local/backend.py
@@ -841,9 +841,9 @@ class Local(base.BaseBackend):
             except exceptions.NotFound:
                 in_tuple = self._db.get(schemas.Type.CompositeTraintuple, key=model_key)
                 in_models.append({
-                    "key": in_tuple.out_head_model.out_model.key,
-                    "checksum": in_tuple.out_head_model.out_model.checksum,
-                    "storage_address": in_tuple.out_head_model.out_model.storage_address,
+                    "key": in_tuple.out_trunk_model.out_model.key,
+                    "checksum": in_tuple.out_trunk_model.out_model.checksum,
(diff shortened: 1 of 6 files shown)


substra/sdk/backends/local/backend.py
```diff
@@ -29,9 +31,17 @@
 
 class Local(base.BaseBackend):
     def __init__(self, backend, *args, **kwargs):
+        self._support_chainkeys = os.getenv("CHAINKEYS_ENABLED", False)
```
jmorel, line 34: I think we should add strtobool around the os.getenv so that `export CHAINKEYS_ENABLED=False` isn't interpreted as a true value here.

jmorel (changes requested):

substra/sdk/backends/local/backend.py
Esadruhn: That's a good catch, I did not know that function

jmorel (approved):

jmorel (approved):

PR #241, opened 2020-11-03
Rename a few attributes

AurelienGasser:
Follow-up to https://github.com/SubstraFoundation/substra/pull/235

Companion PRs:

- https://github.com/SubstraFoundation/substra-backend/pull/345
- https://github.com/SubstraFoundation/substra-chaincode/pull/129
- https://github.com/SubstraFoundation/substra-tests/pull/158

Rename:

| Before | After |
|---|----|
| `compute_plan_id` | `compute_plan_key` (or simply `key`) |
|  `hash` | `checksum` |
| `data_manager.keys` | `data_manager.data_sample_keys` |

### DIFF
--- a/examples/compute_plan/scripts/add_compute_plan.py
+++ b/examples/compute_plan/scripts/add_compute_plan.py
@@ -79,9 +79,9 @@ compute_plan = client.add_compute_plan({
     'composite_traintuples': [],
     'aggregatetuples': [],
 })
-compute_plan_id = compute_plan.compute_plan_id
+key = compute_plan.key
 
 with open(compute_plan_keys_path, 'w') as f:
     json.dump(compute_plan.dict(exclude_none=False, by_alias=True), f, indent=2)
 
-print(f'Compute plan keys have been saved to {os.path.abspath(compute_plan_keys_path)}')
+print(f'Compute Plan keys have been saved to {os.path.abspath(compute_plan_keys_path)}')
--- a/examples/compute_plan/scripts/add_compute_plan.py
+++ b/examples/compute_plan/scripts/add_compute_plan.py
@@ -79,7 +79,6 @@ compute_plan = client.add_compute_plan({
     'composite_traintuples': [],
     'aggregatetuples': [],
 })
-key = compute_plan.key
 
 with open(compute_plan_keys_path, 'w') as f:
     json.dump(compute_plan.dict(exclude_none=False, by_alias=True), f, indent=2)
--- a/references/sdk.md
+++ b/references/sdk.md
@@ -26,10 +26,10 @@ In debug mode, new assets are created locally but can access assets from
 the deployed Substra platform. The platform is in read-only mode.
 Defaults to False.
 ## temp_directory
-_This is a property._  
+_This is a property._
 Temporary directory for storing assets in debug mode.
         Deleted when the client is deleted.
-        
+
 ## add_aggregate_algo
 ```python
 add_aggregate_algo(self, data: Union[dict, substra.sdk.schemas.AggregateAlgoSpec]) -> str
@@ -227,7 +227,7 @@ keys as specified in [schemas.TraintupleSpec](sdk_schemas.md#TraintupleSpec).
  - `str`: Key of the asset
 ## cancel_compute_plan
 ```python
-cancel_compute_plan(self, compute_plan_id: str) -> substra.sdk.models.ComputePlan
+cancel_compute_plan(self, key: str) -> substra.sdk.models.ComputePlan
 ```
 
 Cancel execution of compute plan, the returned object is described
@@ -501,10 +501,10 @@ in the [models.Traintuple](sdk_models.md#Traintuple) model
 login(self, username, password)
 ```
 
-Login to a remote server. 
+Login to a remote server.
 ## update_compute_plan
 ```python
-update_compute_plan(self, compute_plan_id: str, data: Union[dict, substra.sdk.schemas.UpdateComputePlanSpec], auto_batching: bool = True, batch_size: int = 20) -> substra.sdk.models.ComputePlan
+update_compute_plan(self, compute_plan_key: str, data: Union[dict, substra.sdk.schemas.UpdateComputePlanSpec], auto_batching: bool = True, batch_size: int = 20) -> substra.sdk.models.ComputePlan
 ```
 
 Update compute plan.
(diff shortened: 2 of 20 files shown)


AurelienGasser:
`make test`: :heavy_check_mark: 

`make test-local`: :heavy_check_mark:

examples/compute_plan/scripts/add_compute_plan.py
```diff
@@ -79,9 +79,9 @@
     'composite_traintuples': [],
     'aggregatetuples': [],
 })
-compute_plan_id = compute_plan.compute_plan_id
+key = compute_plan.key
```
jmorel, line 82: It looks like this variable isn't used at all, you could remove it entirely.

references/sdk.md
```diff
@@ -504,15 +504,15 @@ login(self, username, password)
...
+update_compute_plan(self, key: str, data: Union[dict, substra.sdk.schemas.UpdateComputePlanSpec], auto_batching: bool = True, batch_size: int = 20) -> substra.sdk.models.ComputePlan
 ```
 
 Update compute plan.
 As specified in the data dict structure, output trunk models of composite
 traintuples cannot be made public.
 
 **Arguments:**
- - `compute_plan_id (str, required)`: Id of the compute plan
+ - `key (str, required)`: Id of the compute plan
```
jmorel, line 515: Could you update the arg description as well so that there is no more mention of "ID".

substra/cli/printers.py
```diff
@@ -264,7 +264,7 @@ class BaseAlgoPrinter(AssetPrinter):
 class ComputePlanPrinter(AssetPrinter):
     asset_name = 'compute_plan'
 
-    key_field = Field('Compute plan ID', 'compute_plan_id')
+    key_field = Field('Compute Plan key', 'key')
```
jmorel, line 267: You could remove this line entirely, it would fallback onto the `key_field = Field('Key', 'key')` (similar to what is done for other fields)

substra/sdk/backends/local/backend.py
```diff
@@ -254,7 +254,7 @@ def __format_for_leaderboard(self, testtuple):
         algo = self._db.get(traintuple.algo_type, traintuple.algo.key)
         return {
             'algo': {
-                'hash': algo.key,
+                'key': algo.key,
```
jmorel, line 257: shouldn't there be a checksu in this object since there also is a storage_address ?

substra/sdk/models.py
```diff
@@ -109,12 +109,12 @@ class _ObjectiveDataset(schemas._PydanticConfig):
...
     """Metric associated to a testtuple or objective"""
     name: Optional[str]
-    hash_: str = pydantic.Field(..., alias="hash")
+    checksum: str
     storage_address: UriPath
 
     @property
     def key(self):
-        return self.hash_
+        return self.checksum
```
jmorel, line 117: Is this still true? I was under the impression that keys and checksum were now always distinct.

tests/test_cli.py
```diff
@@ -297,7 +297,7 @@ def test_command_add_data_sample_already_exists(workdir, mocker):
     ('traintuple', 'key', models.Traintuple),
     ('aggregatetuple', 'key', models.Aggregatetuple),
     ('composite_traintuple', 'key', models.CompositeTraintuple),
-    ('compute_plan', 'compute_plan_id', models.ComputePlan),
+    ('compute_plan', 'key', models.ComputePlan),
```
jmorel, line 300: The second arg value in these parametrized values is always "key" now, maybe you could remove it entirely?

jmorel (changes requested):

AurelienGasser:
@jmorel @Kelvin-M  Thank you both for the reviews!

I'd like to first address a comment by @Kelvin-M, because it might impact some other comments/fixes.

Kelvin's comment is about the inconsistency in using "key" in some places and "compute_plan_key" in other places. 

Here are 2 interesting places where I made the change from "compute_plan_key" to "key":

1. update compute plan
2. cancel compute plan

The rationale for the change is that we're clearly in the context of a compute plan, so "compute_plan_key" is redundant. However I myself am unsure about this change because "compute_plan_key" might be clearer.

What do you guys think? Any opinion as to whether we should use "key" or "compute_plan_key"?

Note : I'm excluding the case of the asset "compute plan" retuned by the chaincode in this conversation. The asset now has a "key" property instead of previously "compute plan id". I don't think that part is controversial, although feel free to argue otherwise if you feel strongly about it.

jmorel:
I support using key in as arg name for update_Compute_plan and cancel_compute_plan, this is perfectly in line with all other method in the Client class.

AurelienGasser:
@Kelvin-M @jmorel Great thanks! I'll proceed and address the other comments.

substra/sdk/models.py
AurelienGasser: Good catch! 

A `Metric` doesn't have a `key`: only the hosting `Objective` does.

I removed the property.

tests/test_cli.py
```diff
@@ -114,7 +114,7 @@ def test_command_login(workdir, mocker):
...
```
jmorel, line 117: As for the `test_command_get `, the key_field parameter is now useless, could you remove it? I'm sorry I forgot to include this comment in my previous review.

jmorel (changes requested):

tests/test_cli.py
AurelienGasser: Fixed in https://github.com/SubstraFoundation/substra/pull/241/commits/7071f4b4f636edc6646124d22d4390062434ee93

Kelvin-M (approved):

substra/sdk/backends/local/backend.py
```diff
@@ -254,7 +254,8 @@ def __format_for_leaderboard(self, testtuple):
...
+                'checksum': algo.checksum,
```
Esadruhn, line 258: algo has no 'checksum' field. This is for the leaderboard, so it returns a subset of the object, and only these fields are printed: https://github.com/SubstraFoundation/substra/blob/master/substra/cli/printers.py#L535

(this function is not tested, whether in debug or normal mode)

Esadruhn (commented):
👍

Esadruhn (dismissed):

inalgnu (approved):

jmorel (approved):

Esadruhn (approved):

PR #248, opened 2020-11-30
Update Substra logo

natct10:
Hi, this PR just updates the Substra logo used on the README with the new design!
Please let me know if the size or format needs to be updated.


jmorel (approved):

jmorel:
BTW do you know when the new logo will be available as a proper svg for integration into the frontend?

Kelvin-M:
We should preserve the same extension for the logo no ? `.svg` instead of using a `.png` file

natct10:
> We should preserve the same extension for the logo no ? `.svg` instead of using a `.png` file

Agreed, but I am not sure that we have this one in `svg`, @ClementMayer do you about this?

jmorel:
This SVG isn't exactly a SVG and that's why I didn't include it in the frontend yet: it a SVG that includes a JPEG inside, which makes it lose the interesting properties of SVG. It fine for this repo, but not for the frontend.

Kelvin-M (dismissed):

jmorel:
Could you update the readme so that the logo isn't gigantic when displayed on github?

jmorel (changes requested):
Could you update the readme so that the logo isn't gigantic when displayed on github?

natct10:
> Could you update the readme so that the logo isn't gigantic when displayed on github?

I know! I was looking for a way to set a size in markdown or to resize the file

Esadruhn (commented):
For the size in the README, does using HTML work ?

`drawing`

natct10:
> For the size in the README, does using HTML work ?
> 
> `drawing`

Yes it should work, I'll try this, thank you @Esadruhn

README.md
```diff
@@ -1,4 +1,4 @@
-# ![Substra](./substra-logo.svg)
+<img src="substra-logo.svg" alt="drawing" width="200"/>
```
jmorel, line 1: Could you add the `<h1> ` around the logo again? And change the alt from "drawing" to "substra" ? Thanks!

jmorel (changes requested):

README.md
natct10: Sure, thank you for this insight @jmorel!
Done

jmorel (approved):

PR #249, opened 2020-12-08
Debug mode - multi nodes - create one local folder per node per compute plan

Esadruhn:
In a 'real' setup, when executing a compute plan on several nodes, there is a local folder for the compute plan on each node, shared by all tasks of the CP that are executed on that node.

- In debug mode, there was one folder per compute plan. This MR makes it so that there is one shared folder per compute plan per node.

- Also added a factory to test the debug mode without substra-tests.
- Fixed TraintupleSpec `from_compute_plan` method in case `in_models_ids` is None

### DIFF
--- a/tests/sdk/test_debug.py
+++ b/tests/sdk/test_debug.py
@@ -23,15 +23,13 @@ def test_client_multi_nodes_cp(asset_factory):
     """Assert that there is one CP local folder per node"""
     client = substra.Client(debug=True)
 
-    dataset_query = asset_factory.create_dataset()
-    dataset_query.metadata = {substra.DEBUG_OWNER: 'owner_1'}
-    dataset_key = client.add_dataset(dataset_query)
+    dataset_query = asset_factory.create_dataset(metadata={substra.DEBUG_OWNER: 'owner_1'})
+    dataset_1_key = client.add_dataset(dataset_query)
 
-    dataset_2_query = asset_factory.create_dataset()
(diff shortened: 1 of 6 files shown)


tests/sdk/test_debug.py
```diff
@@ -15,3 +17,49 @@ def test_client_multi_nodes_dataset(dataset_query):
...
     asset = client.get_dataset(key)
     assert asset.owner == 'owner_1'
+
+
+def test_client_multi_nodes_cp(asset_factory):
+    """Assert that there is one CP local folder per node"""
+    client = substra.Client(debug=True)
+
+    dataset_query = asset_factory.create_dataset()
+    dataset_query.metadata = {substra.DEBUG_OWNER: 'owner_1'}
```
jmorel, line 27: You could set the metadata directly through the create_dataset method.

tests/sdk/test_debug.py
```diff
@@ -15,3 +17,49 @@ def test_client_multi_nodes_dataset(dataset_query):
...
+    dataset_key = client.add_dataset(dataset_query)
```
jmorel, line 28: Could you rename the variable dataset_1_key (similar to what you've done with dataset_2_key) ?

tests/sdk/test_debug.py
jmorel, line 27: You could set the metadata directly through the create_dataset method.

jmorel (approved):

PR #252, opened 2021-01-06
Fix update CP with schema as input

Esadruhn:
### Description

Call the client function `update_compute_plan` with `data` of type schemas.UpdateComputePlanSpec fails.  
trying with `data` of type dict works

### Changes

Fixed the bug and added a test

### DIFF
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -630,7 +630,7 @@ class Client(object):
             models.ComputePlan: updated compute plan, as described in the
             [models.ComputePlan](sdk_models.md#ComputePlan) model
         """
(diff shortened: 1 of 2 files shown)


jmorel (commented):
I'm fine with the changes, but the CI isn't passing. Could you have a look at it?

Esadruhn:
> I'm fine with the changes, but the CI isn't passing. Could you have a look at it?

Sorry, the CI fails because of there are too many calls to Docker: `docker.errors.BuildError: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading`

Who has the right to relaunch it ?

jmorel (approved):

PR #257, opened 2021-02-16
debug: move the chainkeys to local-worker

Esadruhn:
In debug mode, when the chainkeys are enabled, they are created in `~/.substra_chainkeys`

What this PR changes:
- the folder `~/.substra_chainkeys` is now `local-worker/chainkeys`

That way, in debug mode everything is created inside the `local-worker` directory.


Esadruhn:
NB: this would require a change in any code that generates chainkeys

samlesu (approved):

PR #269, opened 2021-05-11
DOC Update to README instructions for titanic

maikia:
A couple of changes to the docs to keep up with  the most recent versions
and ravel y to avoid the warning

Perhaps the substra-docs could also be updated accordingly? 
ie, for example: https://github.com/SubstraFoundation/substra-documentation/blob/master/docs/_sources/demo.md.txt

CC: @Esadruhn

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -27,7 +27,7 @@ jobs:
           export FORCE_ROOTLESS_INSTALL=1
           curl -fsSL https://get.docker.com/rootless | sh
       - name: Install tools
-        run: pip install flake8 wheel docstring-parser
+        run: pip install flake8 wheel docstring-parser==0.7.3
       - name: Lint
         run: flake8 substra
       - name: Install substra
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -40,6 +40,7 @@ jobs:
       - name: Generate and validate CLI documentation
         run: |
           python bin/generate_cli_documentation.py --output-path references/cli.md.tmp
+          diff references/cli.md references/cli.md.tmp
           cmp --silent references/cli.md references/cli.md.tmp
       - name: Generate and validate SDK documentation
         run: |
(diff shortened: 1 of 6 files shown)


examples/titanic/assets/algo_random_forest/algo.py
```diff
@@ -130,7 +130,7 @@ def train(self, X, y, models, rank):
                                                oob_score=True,
                                                random_state=1,
                                                n_jobs=-1)
-        random_forest.fit(X, y)
+        random_forest.fit(X, y.values.ravel())
```
Esadruhn, line 133: was the error due to a specific version of sklearn ? Since I did not reproduce. If that's the case, could you fix the version of sklearn in the Dockerfile ? and maybe add a comment in the README that for these manual tests you need  the right version for the libraries

maikia: It is a warning, not an error in my case:

```DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples,), for example using ravel().```

while `y` is a Pandas dataframe of shape [712 x 1]

I have sklearn v0.2.4, though I am not convinced this is the problem of the version

Still, as it is surprising that you are not seeing it, I will fix the version. It won't harm

Comment (maikia) [reply examples/titanic/assets/algo_random_forest/algo.py:133]: It is a warning, not an error in my case:

DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples,), for example using ravel().

while y is a Pandas dataframe of shape [712 x 1]

I have sklearn v0.2.4, though I am not convinced this is the problem of the version

Still, as it is surprising that you are not seeing it, I fixed the version in the dockerfile. It won't harm

ps. If I run this same algo.py in the debug mode I will see the same warning so it is not only related to the manual tests

**Comment by maikia** [comment]
+ other changes required to keep consistency in face of newer versions of dependencies and make CI pass

File: .github/workflows/python.yml [hunk 2]
```diff
@@ -40,12 +40,16 @@ jobs:
       - name: Generate and validate CLI documentation
         run: |
           python bin/generate_cli_documentation.py --output-path references/cli.md.tmp
+          diff references/cli.md references/cli.md.tmp
```
Esadruhn, line 43: Please remove the diff

maikia:
We decided not to support this documentation for the moment

PR #281, opened 2022-09-12
feat: set the default batch size to 400 to make the compute plan submission faster

Esadruhn:
Patch - include the default batch size update from this commit: https://github.com/Substra/substra/commit/4d9233b000fc34bc64f58b7e3823d66c5aed3fd9
into the release 0.29.2

--> Changing it to 400 as we get timeouts with 450

EDIT: the question is actually more complex than that, discussing it in an issue here: https://github.com/Substra/substra/issues/282

### DIFF
--- a/substra/__version__.py
+++ b/substra/__version__.py
@@ -12,4 +12,4 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "0.29.1"
+__version__ = "0.29.2"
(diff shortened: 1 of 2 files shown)


Esadruhn:
size_of_grpc_packet = const_value * number_of_samples_per_task * number_of_tasks_per_batch
size_of_grpc_packet is capped

we get an error with a CP of 451 tasks, 400 data samples per task

PR #314, opened 2022-10-17
chore: clean test_debug.py

ThibaultFy:
Signed-off-by: ThibaultFy <50656860+ThibaultFy@users.noreply.github.com>

### DIFF
--- a/substra/sdk/backends/local/__init__.py
+++ b/substra/sdk/backends/local/__init__.py
@@ -1,3 +1,0 @@
-from substra.sdk.backends.local.backend import LOCAL_WORKER_DIR
-
-__all__ = ["LOCAL_WORKER_DIR"]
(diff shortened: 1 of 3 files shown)


substra/sdk/backends/local/__init__.py
```diff
@@ -0,0 +1,3 @@
+from substra.sdk.backends.local.backend import LOCAL_WORKER_DIR
+
+__all__ = ["LOCAL_WORKER_DIR"]
```
Esadruhn, line 3: not needed, you can get the dir path with `client.temp_directory`

substra/sdk/backends/local/__init__.py
```diff
@@ -0,0 +1,3 @@
+from substra.sdk.backends.local.backend import LOCAL_WORKER_DIR
```
Esadruhn, line 1: please update the changelog

PR #353, opened 2023-04-17
feat!: client configuration with environment variables or configuration files

SdgJlbl:
## Summary
This PR add the option to configure a Substra client using a configuration file or environment variables. 

The priority order is values given in code (override everything else) > values in environment variables > values in configuration file > default values. 

If both a token and a pair username / password are provided (by any means), the token will always be used. If a token is not provided, one will be generated from the username / password if appropriate (remote or hybrid client). 
It is also possible to regenerate a token by an explicit call to the `login` method.

This PR can be reviewed commit by commit.

## Notes
It's a breaking change because the default (client instantiated with no parameter) will be a local subprocess client instead of a remote client. This decision was taken because the subprocess mode requires no additional configuration, whereas a remote client requires at least an url and a means of authenticating. 

During the implementation, the following decisions have been taken:
- This PR adds a `name` argument to the `Client` class: it is required to be able to match other configurations values (url, username, ...) to a given client when we define several clients in a same script (which is almost all the time in Substra).
- It does NOT use [Pydantic Settings](https://docs.pydantic.dev/usage/settings/): Pydantic has an opinionated, if reasonable, take on settings, and it does not match exactly our needs (in particular regarding the fact that we need dynamical nested configuration for defining several clients at the same time); rather than using hacks to get Pydantic to do what we need, I decided to start from scratch.
- All the configuration occurs in the `__init__`, and not as a `classmethod` that would return an object from a file or a set of variable environments: the configuration options should be seamless from the end user perspective, and switching partially to one option or the other should encourage best security practices (i.e. never storing a password in code or in a file, but only in an env var).
- If a token is provided, it will used over other means of authenticating. 
- The option to create a client from [profiles files created from the CLI](https://github.com/Substra/substra/blob/117a9158dd5abf188c53b9d4553b1ace15a25596/substra/sdk/config.py) is kept as is; it has not been removed, but it also hasn't been reused nor unified with the current work done in this PR. The status of the CLI is still under question, so it was thought best to keep both separated.

Any of the decision above can of course be amended depending on the feedback.

## Open questions
1. Where to we want to convert values to their appropriate type? 
   a. When they are read, using a dict matching each value to its correct type
   b. In the `__init__`, when the values are assigned to attributes
2. Do we want explicit assignation, allowing for finer control on attributes settings, or do we prefer fewer code lines with some "magic" using `setattr`?

## Remaining TODO
- [x] Update doctrings
- [x] Add logging to give the user information about which source was used for configuring each parameter
- [x] Add unit tests
- [x] Check where the organisation name / id is set in the Remote Client and see if we want to unify both usage
- [x] Update user documentation
- [x] Update tutorials 

- [x] If necessary, the [changelog](https://github.com/Substra/substra/blob/main/CHANGELOG.md) has been updated
- [x] Docs have been added / updated (for bug fixes / features)
  - [x] [substra-tests](https://github.com/Substra/substra-tests) https://github.com/Substra/substra-tests/pull/253
  - [x] [substrafl](https://github.com/Substra/substrafl) https://github.com/Substra/substrafl/pull/119
  - [x] [substra-documentation](https://github.com/Substra/substra-documentation) https://github.com/Substra/substra-documentation/pull/305

### DIFF
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -291,11 +291,11 @@ class Client:
             and self._token is None
             and not (config_dict["username"] is None or config_dict["password"] is None)
         ):
-            self._token = self.login(config_dict["username"].value, config_dict["password"].value)
             logger.info(
-                f"No token provided, getting one using username set in {config_dict['username'].origin}"
+                f"No token provided, getting one using username set in {config_dict['username'].origin} "
                 f"and password set in {config_dict['password'].origin}"
             )
+            self._token = self.login(config_dict["username"].value, config_dict["password"].value)
 
     def _get_backend(self, backend_type: schemas.BackendType):
         # Three possibilities:
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -27,6 +27,7 @@ logger = logging.getLogger(__name__)
 
 DEFAULT_RETRY_TIMEOUT = 5 * 60
 DEFAULT_BATCH_SIZE = 500
+DEFAULT_BACKEND_TYPE = schemas.BackendType.LOCAL_SUBPROCESS
 
 # Temporary output identifiers, to be removed once user can pass its own identifiers
 MODEL_OUTPUT_IDENTIFIER = "model"
@@ -174,18 +175,21 @@ class Client:
     """Create a client.
     Defaults to a subprocess client, suitable for development purpose.
     Configuration can be passed in the code, through environment variables or through a configuration file.
+    The order of precedence is: values defined by the user in the code, environment variables, values read from the
+    configuration file. If the attribute is not set, the value returned is None (and the origin is set to "default").
+    In order to use configuration values not explicitly defined in the code, the parameter `configuration_name` must
+    not be None.
 
     Args:
         configuration_name (str, optional): Name of the configuration.
             Used to load relevant environment variables and select the right dictionary in the configuration file.
-            If None, configuration values will only be read from the code, or default will be used.
             Defaults to None.
         configuration_file (path, optional): Path to te configuration file.
             `configuration_name` must be defined too.
             Defaults to None.
         url (str, optional): URL of the Substra platform.
-            Mandatory to connect to a Substra platform. If no URL is given debug must be True and all
-            assets must be created locally.
+            Mandatory to connect to a Substra platform. If no URL is given, all assets are created locally.
+            Can be set to "" to remove any previously defined URL (in a configuration file or environment variable).
             Defaults to None.
         token (str, optional): Token to authenticate to the Substra platform.
             If no token is given, and a `username`/ `password` pair is provided,  the Client will try to authenticate
@@ -197,7 +201,7 @@ class Client:
             Not stored.
             Defaults to None.
         password (str, optional): Password to authenticate to the Substra platform.
-            Used in conjunction with an username to generate a token if not given, using the `login` function.
+            Used in conjunction with a username to generate a token if not given, using the `login` function.
             Not stored.
             Defaults to None.
         retry_timeout (int, optional): Number of seconds before attempting a retry call in case of timeout.
@@ -229,7 +233,9 @@ class Client:
         backend_type: Optional[schemas.BackendType] = None,
     ):
         if configuration_file and not configuration_name:
-            logger.info("Configuration file ignored because no `configuration_name` was given.")
+            raise exceptions.BadConfiguration(
+                "Configuration file cannot be used because no `configuration_name` was given."
+            )
 
         code_values = {
             "url": url,
@@ -245,13 +251,18 @@ class Client:
         )
         if config_dict["backend_type"].value is None:
             logger.info("No backend type specified, defaulting to subprocess")
-            backend_type = schemas.BackendType.LOCAL_SUBPROCESS
+            backend_type = DEFAULT_BACKEND_TYPE
         else:
             backend_type = config_dict["backend_type"].value
             logger.info(f"Backend type set to {backend_type}, " f"as defined in {config_dict['backend_type'].origin}")
 
         self._url = config_dict["url"].value
-        logger.info(f"The parameter `url` has been set to {self._url} " f"as defined in {config_dict['url'].origin} ")
+        if self._url:
+            logger.info(
+                f"The parameter `url` has been set to {self._url} " f"as defined in {config_dict['url'].origin} "
+            )
+        else:
+            logger.info("No URL has been given, all assets will be created locally.")
 
         self._token = config_dict["token"].value
         if self._token:
@@ -321,7 +332,7 @@ class Client:
 
     @property
     def temp_directory(self):
-        """Temporary directory for storing assets in debug mode.
+        """Temporary directory for storing assets in local mode.
         Deleted when the client is deleted.
         """
         if isinstance(self._backend, backends.Local):
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -1,12 +1,19 @@
+import dataclasses
 import functools
 import logging
 import os
 import pathlib
 import time
 import uuid
+from typing import Any
+from typing import Dict
 from typing import List
+from typing import Literal
 from typing import Optional
 from typing import Union
+
+import yaml
+from slugify import slugify
 
 from substra.sdk import backends
 from substra.sdk import config as cfg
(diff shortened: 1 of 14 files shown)


Owlfred:
End to end tests: :hourglass_flowing_sand: **STARTING**

substra/sdk/client.py
```diff
@@ -76,21 +143,59 @@ class Client:
...
+        }
+        config_dict = get_client_configuration(
+            client_name=name, config_file=configuration_file, code_values=code_values
+        )
+
+        if name is None:
+            self.name = uuid.uuid4()
+        else:
+            self.name = name
+
```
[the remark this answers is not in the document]
SdgJlbl: No check is performed for now. 
I'm not sure about what is the best way to handle this check, given that we want to check _across_ clients...

substra/sdk/client.py
```diff
@@ -76,21 +143,59 @@ class Client:
             and tasks are executed locally.
     """
 
+    # TODO update docstring
```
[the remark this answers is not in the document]
SdgJlbl: > Do you plan to propose a "config_file" template ?

maybe in the user doc?

substra/sdk/client.py
```diff
@@ -49,6 +57,65 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def _get_env_var(attribute_name, client_name):
+    return f"SUBSTRA_{slugify(client_name).upper()}_{slugify(attribute_name).upper()}"
```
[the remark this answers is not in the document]
SdgJlbl: good catch

SdgJlbl:
> Also, it seems that your implementation allows "multiple" source. Like some of the config can come from the user, and the rest from a config file for instance... Just raising the point. Can it bring some confusion from a user point of view ? (no opinion for now...)

Yes, the goal is for example to have most of your configuration in a config file, and your secrets (password or token) stored in env var.

Owlfred:
End to end tests: :hourglass_flowing_sand: **STARTING**

substra/sdk/client.py
```diff
@@ -49,26 +56,158 @@ def wrapper(*args, **kwargs):
...
+    """Create a client.
+    Defaults to a subprocess client, suitable for development purpose.
+    Configuration can be passed in the code, through environment variables or through a configuration file.
 
     Args:
-        url (str, optional): URL of the Substra platform. Mandatory
-            to connect to a Substra platform. If no URL is given debug must be True and all
+        configuration_name (str, optional): Name of the configuration.
+            Used to load relevant environment variables and select the right dictionary in the configuration file.
+            If None, configuration values will only be read from the code, or default will be used.
```
ThibaultFy, line 181: It could also be read from env variable ?
Maybe it could be easier to explain the order of priority once and for all in the intro text of the docsting, but not in the args.

substra/sdk/client.py
```diff
@@ -78,19 +217,74 @@ class Client:
...
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        retry_timeout: Optional[int] = None,
+        insecure: Optional[bool] = None,
+        backend_type: Optional[schemas.BackendType] = None,
     ):
-        self._retry_timeout = retry_timeout
-        self._token = token
+        if configuration_file and not configuration_name:
+            logger.info("Configuration file ignored because no `configuration_name` was given.")
```
ThibaultFy, line 232: Is there a use case where it could be interesting for a user to pass a file without a config name ? If not, might be better to raise an user friendly error here ?

substra/sdk/client.py
```diff
@@ -49,26 +56,158 @@ def wrapper(*args, **kwargs):
...
+
+    # Some values are boolean that might be set to False, so we check explicitly "is not None"
+    # Note that explicitly passing "None" in the code is equivalent to omitting it.
+    # (code_values will actually always contain all the keys, with None values for keys that were omitted)
+    if code_values.get(attribute_name) is not None:
+        return SettingValue(value=code_values[attribute_name], origin="code")
+
+    # if client_name is not given, we cannot look through env var and config files
+    if not configuration_name:
+        return SettingValue(value=None, origin="default")
```
ThibaultFy, line 112: If we raise the error, we can remove this part

substra/sdk/client.py
```diff
@@ -78,19 +217,74 @@ class Client:
...
+        )
+        if config_dict["backend_type"].value is None:
+            logger.info("No backend type specified, defaulting to subprocess")
+            backend_type = schemas.BackendType.LOCAL_SUBPROCESS
+        else:
+            backend_type = config_dict["backend_type"].value
+            logger.info(f"Backend type set to {backend_type}, " f"as defined in {config_dict['backend_type'].origin}")
+
+        self._url = config_dict["url"].value
+        logger.info(f"The parameter `url` has been set to {self._url} " f"as defined in {config_dict['url'].origin} ")
```
ThibaultFy, line 254: This log will be weird in subprocess ? As url is always to None

substra/sdk/client.py
```diff
@@ -78,19 +217,74 @@ class Client:
...
+            "retry_timeout": retry_timeout,
+            "insecure": insecure,
+            "backend_type": backend_type,
+        }
+        config_dict = get_client_configuration(
+            configuration_name=configuration_name, config_file=configuration_file, code_values=code_values
+        )
+        if config_dict["backend_type"].value is None:
+            logger.info("No backend type specified, defaulting to subprocess")
+            backend_type = schemas.BackendType.LOCAL_SUBPROCESS
```
ThibaultFy, line 248: What do you think about storing `schemas.BackendType.LOCAL_SUBPROCESS` on top of the file in a `DEFAULT_BACKEND_TYPE` to match the logic of `DEFAULT_RETRY_TIMEOUT` for instance ?

tests/sdk/test_client.py
```diff
@@ -0,0 +1,188 @@
...
+
+def test_client_configuration_from_env_var(mocker, monkeypatch):
+    mocker.patch("substra.sdk.Client.login", side_effect=stub_login)
+    monkeypatch.setenv("SUBSTRA_TOTO_BACKEND_TYPE", "remote")
+    monkeypatch.setenv("SUBSTRA_TOTO_URL", "toto-org.com")
+    monkeypatch.setenv("SUBSTRA_TOTO_USERNAME", "env_var_username")
+    monkeypatch.setenv("SUBSTRA_TOTO_PASSWORD", "env_var_password")
+    monkeypatch.setenv("SUBSTRA_TOTO_RETRY_TIMEOUT", "42")
+    monkeypatch.setenv("SUBSTRA_TOTO_INSECURE", "true")
+    client = Client(configuration_name="toto")
```
ThibaultFy, line 116: What does the configuration_name in that case ?

ThibaultFy (commented):
Thanks a lot for the PR ! :)
Seems really great and well tested. I'll try it on the doc example to see the log rendering mostly.
And we'll need to relaunch the e2e test when fixed :)

Also, when doing a PR in Substra SDK and modifying some docstring, you need to run "make doc" to update it (and commit the changes).

substra/sdk/client.py
SdgJlbl: I don't think we can because it's still acceptable to configure a `Client` entirely from code (and so without any configuration_name). 
I think it's a bit clearer to quit at that point.

tests/sdk/test_client.py
SdgJlbl: missing a verb in this question I think 😅

ThibaultFy: Woooops ^^" What does the configuration_name does in that case ?

ThibaultFy: I mean what the expected behaviour of passing a config name without a config file?

substra/sdk/client.py
ThibaultFy: I know that in SubstraFL we have the really explicit reflex on error, so it would be `ConfigurationNameNotFoundError`, smthg like that

tests/sdk/test_client.py
SdgJlbl: It is used to build the env var names, such as `SUBSTRA_TOTO_PASSWORD`. 
I was wondering if it was necessary to add an how-to guide for this feature, but given the many questions you have, I guess mine is answered 😅

Owlfred:
End to end tests: :hourglass_flowing_sand: **STARTING**

substra/sdk/client.py
```diff
@@ -78,19 +221,81 @@ class Client:
 
     def __init__(
         self,
+        *,
+        configuration_name: Optional[str] = None,
```
RomainGoussault, line 225: @SdgJlbl Why naming it `configuration_name` and not `name`? or `client_name`? 🤔

SdgJlbl: That is the conclusion of the discussion we had after the daily yesterday. This name is only used for the configuration of the `Client`, and we don't want it to be confused with the organisation name (that you can get from `client.organization_info()`). 
We cannot use the organisation name, that is a characteristic of the backend, because we need to be able to connect to the backend to have access to it (chicken and egg problem). 
And we cannot set the organisation name to this value, because it's set in the backend when the backend is created (so earlier than the creation of the client). 
So we thought that an appropriate option would to be making very clear that this name is only used for configuring the client, and not stored afterwards to avoid the confusion of having two similar but not synchronised concepts.

tests/sdk/test_client.py
ThibaultFy: Aaaaah ok ok I understand. Documentation might be welcome yes ahah :)

substra/sdk/client.py
RomainGoussault: Thanks for the detailed explanation. This makes sense and rules out the `name` naming that could be confusing with the organisation name.

I don't have a strong opinion on this, but I think `client_name`, really dissociates from `organization_name` and is less confusing / new than `configuration_name`. As there might be confusion between `configuration_name` and `configuration_file`.

Owlfred:
End to end tests: :hourglass_flowing_sand: **STARTING**

substra/sdk/client.py
```diff
@@ -78,19 +221,79 @@ class Client:
...
+        username: Optional[str] = None,
+        password: Optional[str] = None,
+        retry_timeout: Optional[int] = None,
+        insecure: Optional[bool] = None,
+        backend_type: Optional[schemas.BackendType] = None,
     ):
-        self._retry_timeout = retry_timeout
-        self._token = token
+        if configuration_file and not client_name:
+            raise exceptions.BadConfiguration("Configuration file cannot be used because no `client_name` was given.")
```
ThibaultFy, line 236: Not a fan of this error name. Custom errors should at least end by `*Error`. 
`ConfigurationInfoError` or `ClientNameError` would be better IMO.

ThibaultFy (commented):
Small comment on the error name and will be good for me :) 
You need to re-run the make doc command due to the `client_name` renaming in docstring.

Owlfred:
End to end tests: :hourglass_flowing_sand: **STARTING**

Owlfred:
End to end tests: :hourglass_flowing_sand: **STARTING**

Owlfred:
End to end tests: :hourglass_flowing_sand: **STARTING**

SdgJlbl:
Passing end-to-end tests https://github.com/Substra/substra-tests/pull/253#issuecomment-1532575012

ThibaultFy (approved):
LGTM, thanks a lot for this feature !! 🥳

PR #357, opened 2023-05-03
chore: update performance view

ThibaultFy:
# Description

Performance asset were unique regarding their compute task key, and the function key they were computed from. Or, we want to have the possibility to have a function that outputs several performance and store them in the same task.

This implies two main things:
   - Performance are now unique regarding the compute task key, the function key AND the output identifier
   - The frontend must serialize regarding the identifier, and not the function key

The PR batch modify both the orchestrator and backend db in order to include the identifier as an additional unique condition.
In the backend, we use the ComputeTaskOutput as primary key directly.

## Companion PR

- https://github.com/Substra/substra-backend/pull/634
- https://github.com/Substra/substra-tests/pull/251 (main)
- https://github.com/Substra/orchestrator/pull/197
- https://github.com/Substra/substra-frontend/pull/194
- https://github.com/Substra/substra/pull/357

## E2E tests
- https://github.com/owkin/substra-ci/actions/runs/4872054802

### DIFF
--- a/tests/datastore.py
+++ b/tests/datastore.py
@@ -582,6 +582,7 @@ COMPUTE_PLAN_PERF = {
                 "key": "e526243f-f51a-4737-9fea-a5d55f4205fe",
                 "name": "accuracy",
             },
+            "identifier": "performance",
             "perf": 0.673,
         },
         {
@@ -602,6 +603,7 @@ COMPUTE_PLAN_PERF = {
                 "key": "e526243f-f51a-4737-9fea-a5d55f4205fe",
                 "name": "accuracy",
             },
+            "identifier": "performance",
             "perf": 0.834,
         },
         {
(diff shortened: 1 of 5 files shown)


tests/datastore.py
```diff
@@ -582,6 +582,7 @@
                 "key": "e526243f-f51a-4737-9fea-a5d55f4205fe",
                 "name": "accuracy",
             },
+            "identifier": "performance",
```
SdgJlbl, line 585: can you add a test with multiple performances?

ThibaultFy: There is already multiple performances here, as it is the request to get all performances of 1 compute plan. The test would be to add a performance with the same compute task key but this not tested in substra as it is a backend feature.

SdgJlbl: (sorry if I'm being slow, it's Friday afternoon)
but then, if we can cover it at unit test level, why don't we cover it at unit test level? End-to-end tests are all good but they are harder to run, slower and less easy to debug.

SdgJlbl (approved):

PR #374, opened 2023-07-22
fix: missing dependency (yaml) in requirements

SdgJlbl:


SdgJlbl:
`pyyaml` is already in the requirements, initial problem with Binder must be coming from somewhere else.

PR #378, opened 2023-07-27
chore: add warning to Client.login

oleobal:
## Summary

This adds a warning to `Client.login` telling you to use tokens from the frontend instead. Also tells you when your session will expire.

I purposefully avoided calling it "deprecated" (since we have no plans to remove it), but maybe we should.

Closes FL-1112

- [ ] If necessary, the [changelog](https://github.com/Substra/substra/blob/main/CHANGELOG.md) has been updated
- [ ] Docs have been added / updated (for bug fixes / features)


substra/sdk/backends/remote/rest_client.py
```diff
@@ -63,14 +63,20 @@ def login(self, username, password):
...
+            token = rj["token"]
+            expires_at = rj["expires_at"]
         except json.decoder.JSONDecodeError:
             # sometimes requests seem to be fine, but the json is not being found
             # this might be if the url seems to be correct (in the syntax)
             # but it's not the right one
             raise exceptions.BadConfiguration(
                 "Unable to get token from json response. " f"Make sure that given url: {self._base_url} is correct"
             )
+        logger.warning(
```
SdgJlbl, line 94: Nitpicking but I feel like there are two distinct pieces of information in that message, that would desserve two logs: 
1. client.login is discouraged
2. expiration of the session

For me, the first one should be at the `info` level, not warning (at least, while the login is not deprecated). It might be also worth it to add a link to the documentation to explain why (I must admit that it's not entirely clear for me, especially since we allow very long-lived tokens, which are basically acting as passwords 🤷 ).
The second point should be an `info`  if the expiration is "in a long time" (to be defined), a warning else.

SdgJlbl: Ok so the last commit addresses most of my comments already 🔮

oleobal: > It might be also worth it to add a link to the documentation to explain why (I must admit that it's not entirely clear for me, especially since we allow very long-lived tokens, which are basically acting as passwords 🤷 )

That is a good point, I'll write up a short documentation page then. Not sure where it should go though 😅 

> The second point should be an info if the expiration is "in a long time" (to be defined), a warning else.

I agree with the reasoning, but like you I don't know what constitutes a long time. It depends on the total possible lifetime of the token and we just don't know that here. The backend knows though.

It'd make sense to have the same warning for UI-generated tokens, but that'd mean having to add a new endpoint in the backend and querying it when `Client` is instanciated with a token.

SdgJlbl: For the documentation, I think it could go in the how-to guide about how to use tokens (having a paragraph or a box about why it is the recommended safest method to use).

SdgJlbl (approved):

PR #381, opened 2023-08-09
feat: add Client.logout and context manager

oleobal:
## Summary

Add a `Client.logout` function, which deletes a token obtained by `Client.login` (if any).

Add a context manager, making the following snippet legal:

```python
with Client(
    client_name="org-1",
    backend_type="remote",
    url="http://substra-backend.org-1.com:8000",
    username="org-1",
    password="p@sswr0d44",
) as client:
    pass
```

Requires https://github.com/Substra/substra-backend/pull/698

Closes FL-1008

- [x] If necessary, the [changelog](https://github.com/Substra/substra/blob/main/CHANGELOG.md) has been updated
- [x] Docs have been added / updated (for bug fixes / features)

### DIFF
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -199,6 +199,12 @@ class Client:
             Defaults to None.
         username (str, optional): Username to authenticate to the Substra platform.
             Used in conjunction with a password to generate a token if not given, using the `login` function.
+
+            If using username/password, you should use a context manager to ensure the session is terminated:
+            ```
+            with Client(username, password) as client:
+               ...
+            ```
             Not stored.
             Defaults to None.
         password (str, optional): Password to authenticate to the Substra platform.
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -387,6 +387,7 @@ class Client:
         if not self._backend:
             raise exceptions.SDKException("No backend found")
         self._backend.logout()
+        self._token = None
 
     @staticmethod
     def _get_spec(asset_type, data):
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -386,8 +386,7 @@ class Client:
         """
         if not self._backend:
             raise exceptions.SDKException("No backend found")
-        if hasattr(self._backend, "logout"):
-            self._backend.logout()
+        self._backend.logout()
 
     @staticmethod
     def _get_spec(asset_type, data):
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -200,7 +200,7 @@ class Client:
         username (str, optional): Username to authenticate to the Substra platform.
             Used in conjunction with a password to generate a token if not given, using the `login` function.
 
-            If using username/password, you should use a context manager to ensure the session is terminated:
+            If using username/password, you should use a context manager to ensure the session terminates as intended:
             ```
             with Client(username, password) as client:
                ...
--- a/substra/sdk/client.py
+++ b/substra/sdk/client.py
@@ -311,6 +311,14 @@ class Client:
             )
             self._token = self.login(config_dict["username"].value, config_dict["password"].value)
 
+    def __enter__(self):
+        # for use in a `with` statement
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        # for use in a `with` statement
+        self.logout()
+
     def _get_backend(self, backend_type: schemas.BackendType):
         # Three possibilities:
         # - remote: get a remote backend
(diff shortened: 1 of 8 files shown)


substra/sdk/client.py
```diff
@@ -311,6 +311,15 @@ def __init__(
...
             self._token = self.login(config_dict["username"].value, config_dict["password"].value)
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.logout()
+
+    def __del__(self):
+        self.logout()
```
oleobal, line 321: Unsure whether context manager is still needed with `__del__` defined, but people online say not to rely on it and using `with` is preferable. However that does mean updating our documentation.

substra/sdk/client.py
```diff
@@ -369,6 +378,17 @@ def login(self, username, password):
...
+    @logit
+    def logout(self) -> None:
+        """
+        Log out from a remote server, if Client.login was used
+        (otherwise, nothing happens)
+        """
+        if not self._backend:
+            raise exceptions.SDKException("No backend found")
+        if hasattr(self._backend, "logout"):
+            self._backend.logout()
```
ThibaultFy, line 390: I would check the client type rather than `hasattr` if the purpose of this is to check that we have a remote client.

substra/sdk/client.py
ThibaultFy: What part of the documentation are you reffering to ?

ThibaultFy (commented):
Thanks for the PR :) 

Concerning the fact that we define a logout function only for the remote backend, I am not 100% convince... There is a login method defined in locals backend too, I think for the hybrid mode. You should take a look at it to ensure what we should do for this mode I think. (cf `sdk/backends/local/dal.py`.

For the tests, you could check that the token is deleted the using the context manager `with` or when deleting the object (`del client`).
This test can be useful to mock minimally https://github.com/Substra/substra/blob/eae5d52ab804c9a4d6794ef299a078790069b253/tests/sdk/test_client.py#L92

substra/sdk/client.py
oleobal: I mean that in the documentation all examples of the form:

```
client = Client(...)
do_stuff(client)
```

should be replaced with:


```python
with Client(...) as client:
    do_stuff(client)
```

substra/sdk/client.py
oleobal: That makes sense, but isn't the purpose at this level just to check whether the client has a `logout` function, regardless of what kind of client it is?

ThibaultFy: Regarding my main comment, I think it would make more sense that all client have a logout, as they all have a login method

substra/sdk/client.py
ThibaultFy: Some scripts (such as the example) need the client for more than a thousand lines, no I would not replace the old version, maybe just indicate that the context manager is also a valid option ?

oleobal: I added it in the `sdk.Client` reference, under the `username` argument.

substra/sdk/client.py
```diff
@@ -199,6 +199,12 @@ class Client:
             Defaults to None.
         username (str, optional): Username to authenticate to the Substra platform.
             Used in conjunction with a password to generate a token if not given, using the `login` function.
+
+            If using username/password, you should use a context manager to ensure the session is terminated:
```
ThibaultFy, line 203:
```suggestion
            If using username/password, you should use a context manager to ensure the session terminates as intended:
```

ThibaultFy (commented):
A few small comments and will be good to me ! do you think we need a run of CI on it before merging ? No strong opinion on it on my side.

oleobal:
@ThibaultFy 
> do you think we need a run of CI on it before merging ? No strong opinion on it on my side.

I [ran one yesterday](https://github.com/Substra/substra-backend/pull/698#issuecomment-1680434880) but let's run a new one just to be sure ^^

ThibaultFy:
> @ThibaultFy
> 
> > do you think we need a run of CI on it before merging ? No strong opinion on it on my side.
> 
> I [ran one yesterday](https://github.com/Substra/substra-backend/pull/698#issuecomment-1680434880) but let's run a new one just to be sure ^^

Ah ok ok :) you're not using the /e2e command ?

oleobal:
> you're not using the /e2e command ?

I am, just from the backend PR instead of this one

ThibaultFy (approved):
LGTM, thanks a lot :)

PR #392, opened 2023-10-12
feat: resolve relative datasample spec path

ThibaultFy:
## Related issue

See slack https://owkin.slack.com/archives/C01UWGW68N8/p1696943277910689 for context
closes FL-1243

## Summary

Resolve `path` and `paths` on datasample spec.

### DIFF
--- a/substra/sdk/schemas.py
+++ b/substra/sdk/schemas.py
@@ -147,7 +147,8 @@ class DataSampleSpec(_Spec):
         return self.paths and len(self.paths) > 0
 
     @pydantic.model_validator(mode="before")
-    def exclusive_paths(cls, values: typing.Any) -> typing.Any:  # noqa: N805
+    @classmethod
+    def exclusive_paths(cls, values: typing.Any) -> typing.Any:
         """Check that one and only one path(s) field is defined."""
         if "paths" in values and "path" in values:
             raise ValueError("'path' and 'paths' fields are exclusive.")
@@ -156,7 +157,8 @@ class DataSampleSpec(_Spec):
         return values
 
     @pydantic.model_validator(mode="before")
-    def resolve_paths(cls, values: typing.Any) -> typing.Any:  # noqa: N805
+    @classmethod
+    def resolve_paths(cls, values: typing.Any) -> typing.Any:
         """Resolve given path is relative."""
         if "paths" in values:
             paths = []
@@ -310,7 +312,8 @@ class FunctionInputSpec(_Spec):
     kind: AssetKind
 
     @pydantic.model_validator(mode="before")
-    def _check_identifiers(cls, values):  # noqa: N805
+    @classmethod
+    def _check_identifiers(cls, values):
         """Checks that the multiplicity and the optionality of a data manager is always set to False"""
         if values["kind"] == AssetKind.data_manager:
             if values["multiple"]:
@@ -344,7 +347,8 @@ class FunctionOutputSpec(_Spec):
     multiple: bool
 
     @pydantic.model_validator(mode="before")
-    def _check_performance(cls, values):  # noqa: N805
+    @classmethod
+    def _check_performance(cls, values):
         """Checks that the performance is always set to False"""
         if values == AssetKind.performance and values["multiple"]:
             raise ValueError("Performance can't be multiple.")
@@ -369,7 +373,8 @@ class FunctionSpec(_Spec):
     type_: typing.ClassVar[Type] = Type.Function
 
     @pydantic.field_validator("inputs")
-    def _check_inputs(cls, v):  # noqa: N805
+    @classmethod
+    def _check_inputs(cls, v):
         inputs = v or []
         identifiers = {value.identifier for value in inputs}
         if len(identifiers) != len(inputs):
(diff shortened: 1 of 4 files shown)


substra/sdk/schemas.py
```diff
@@ -155,6 +155,23 @@ def exclusive_paths(cls, values):  # noqa: N805
             raise ValueError("'path' or 'paths' field must be set.")
         return values
 
+    @pydantic.model_validator(mode="before")
+    def resolve_paths(cls, values):  # noqa: N805
```
thbcmlowk, line 159: No typing on `values`?

ThibaultFy: I just copy paste it from abose. We might add it though, but it's `values` is provided by pydantic so it's not that bad no ? Wdyt ?

substra/sdk/schemas.py
```diff
@@ -155,6 +155,23 @@ def exclusive_paths(cls, values):  # noqa: N805
...
+        """Resolve given path is relative."""
+        if "paths" in values:
+            paths = []
+            for path in values["paths"]:
+                path = pathlib.Path(path)
+                paths.append(path.resolve()) if not path.is_absolute() else paths.append(path)
```
guilhem-barthes, line 165:
```suggestion
                paths.append(path.resolve())
```
Do we need to check `is_absolute()`  as`path.resolve()` will return itself if already a absolute path?

substra/sdk/schemas.py
guilhem-barthes: I would rather use `values: Any): -> Any` or create a custom type (e.g. `ValueType`) and use `values: ValueType): -> ValueType`

guilhem-barthes (commented):
Thanks for your work!

substra/sdk/schemas.py
```diff
@@ -147,14 +147,31 @@ def is_many(self):
         return self.paths and len(self.paths) > 0
 
     @pydantic.model_validator(mode="before")
-    def exclusive_paths(cls, values):  # noqa: N805
+    def exclusive_paths(cls, values: typing.Any) -> typing.Any:  # noqa: N805
```
guilhem-barthes, line 150: can you remove the`# noqa: N805`?

substra/sdk/schemas.py
```diff
@@ -147,14 +147,31 @@ def is_many(self):
...
         """Check that one and only one path(s) field is defined."""
         if "paths" in values and "path" in values:
             raise ValueError("'path' and 'paths' fields are exclusive.")
         if "paths" not in values and "path" not in values:
             raise ValueError("'path' or 'paths' field must be set.")
         return values
 
+    @pydantic.model_validator(mode="before")
+    def resolve_paths(cls, values: typing.Any) -> typing.Any:  # noqa: N805
```
guilhem-barthes, line 159: can you remove the `# noqa: N805`? It should be handled by the typing

substra/sdk/schemas.py
```diff
@@ -147,14 +147,31 @@ def is_many(self):
         return self.paths and len(self.paths) > 0
 
     @pydantic.model_validator(mode="before")
-    def exclusive_paths(cls, values):  # noqa: N805
+    def exclusive_paths(cls, values: typing.Any) -> typing.Any:  # noqa: N805
         """Check that one and only one path(s) field is defined."""
         if "paths" in values and "path" in values:
             raise ValueError("'path' and 'paths' fields are exclusive.")
         if "paths" not in values and "path" not in values:
             raise ValueError("'path' or 'paths' field must be set.")
```
guilhem-barthes, line 155: Could they define both keys but with value `None`?

substra/sdk/schemas.py
ThibaultFy: Nop, its because it does not like that the first arg is not "self" :/

substra/sdk/schemas.py
ThibaultFy: I don't get what you're refering too. Could you provide an example plz ?

substra/sdk/schemas.py
guilhem-barthes: If this method entends to receive a class as first argument, pydantic recommends decorate with `@classmethod` for a proper type checking ([model_validator](https://docs.pydantic.dev/latest/concepts/validators/#model-validators))

ThibaultFy: nice thx :)

substra/sdk/schemas.py
guilhem-barthes: It looks like we allow `None` as value on these fields, therefore if a user explicitely define only one of this key to `None` (e.g. `DataSampleSpec(path=None, data_manager_keys=[])`, the check `"paths" not in values and "path" not in values` will return False (as the key is present, but the value is `None`).

ThibaultFy: Ok I see. I added some tests, and it appears that this case is dealts with in the resolve path, as `None` will raise a `TypeError`. This is not an explicit solution but I added the tests to ensure that this behavior is taken into account. What do you think ?

guilhem-barthes: I don't know if `TypeError` is the clearest for the user, but maybe it is worth opening its own PR to review how we manage this edge case?

ThibaultFy: Yep True... I rm the tests, and +1 to deal with this in a dedicated PR

guilhem-barthes (approved):
Thanks for your work!

PR #416, opened 2024-06-03
chore: release 0.53.0

ThibaultFy:


thbcmlowk (approved):
🐍

substra/__version__.py
```diff
@@ -1 +1 @@
-__version__ = "0.52.0"
+__version__ = "0.53.0"
```
thbcmlowk, line 1: Would it not rather be: 
```suggestion
__version__ = "0.52.1"
```

